PR #14, opened 2024-07-10
Rate limiting tiers tutorial

cloudjumpercat:
First pass at a Rate Limiting Tiers (based on https://docs.konghq.com/hub/kong-inc/rate-limiting-advanced/how-to/) tutorial. I tried using the entity_example Fabian made, as well as a new plugin_example in the same style. I also added an info box to the front matter along with a FAQ just to try those out.

### DIFF
--- a/app/_tutorials/rate-limiting-tiers.md
+++ b/app/_tutorials/rate-limiting-tiers.md
@@ -1,0 +1,103 @@
+---
+title: How to create rate limiting tiers
+related_resources:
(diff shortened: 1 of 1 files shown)


fabianrbz:
@cloudjumpercat this is a great start 🎉 !
a few more things to consider:
* The question/answer as a TLDR; block we have at the top after the title
* Prereqs
* min version in which this works
* tier is it OSS/EE

PR #20, opened 2024-07-16
[don't merge yet] feat: Rate Limiting Plugin Reference Page

cloudjumpercat:
- I added more front matter to the existing Rate Limiting reference page
- I also added an example plugin schema block in Liquid to see if something like that would work (I didn't do any of the backend work to make the schema render because I have no idea what would be done to make that work)
- I can't help but wonder if reference pages like this (that are primarily schema-driven) should be yaml files instead with the `row` and `column` formatting instead. I don't think people other than the docs team will be editing these pages, so I wonder if yaml files would work or be better than markdown since there's less text content.

### DIFF
--- a/app/_kong_plugins/rate-limiting/reference.md
+++ b/app/_kong_plugins/rate-limiting/reference.md
@@ -1,0 +1,54 @@
+---
+title: Rate Limiting configuration
+content_type: reference
+entities: 
+    - service
+    - route
+    - consumer
+    - consumer_group
+    - plugin
+faqs:
+  - q: What is the Rate Limiting plugin used for?
+    a: The Rate Limiting plugin is used to control the rate of requests that clients can make to your services. It helps prevent abuse and ensures fair usage by limiting the number of requests a client can make in a given time period.
+  - q: Can I set different rate limits for different endpoints or services?
+    a: Yes, you can configure rate limits on a per-service or per-route basis by applying the Rate Limiting plugin to specific services or routes in Kong.
+  - q: "How does the `policy` option affect rate limiting?"
+    a: |
+      The `policy` option determines how rate limits are stored and enforced. The `local` policy uses Kong’s in-memory storage, while the `redis` policy uses Redis, which is useful for distributed setups where rate limiting needs to be consistent across multiple Kong data plane nodes.
+related_resources:
+  - text: How to create rate limiting tiers with Rate Limiting Advanced
+    url: https://docs.konghq.com/hub/kong-inc/rate-limiting-advanced/how-to/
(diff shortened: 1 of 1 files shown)


app/_kong_plugins/rate-limiting.md
```diff
@@ -11,11 +17,26 @@ faqs:
...
+
+The advanced version of this plugin, [Rate Limiting Advanced](https://docs.konghq.com/hub/kong-inc/rate-limiting-advanced/), provides the ability to apply [mutliple limits in sliding or fixed windows](https://docs.konghq.com/hub/kong-inc/rate-limiting-advanced/#multi-limits-windows).
+
+<!-- example plugin schema renderer below, ideally this will render on the left of the config examples -->
+
+{% schema %}
+type: plugin
+data:
+  name: rate-limiting
+{% endschema %}
```
cloudjumpercat, line 39: @fabianrbz How can we add the plugin schema so it can be pulled in here?

fabianrbz: We don't have support for that yet. You can just leave an h2 saying something like schema goes here

cloudjumpercat, line 39:
```suggestion
## schema goes here

<!-- example plugin schema renderer below, ideally this will render on the left of the config examples 

{% schema %}
type: plugin
data:
  name: rate-limiting
{% endschema %}
-->
```

app/_kong_plugins/rate-limiting.md
```diff
@@ -11,11 +17,29 @@ faqs:
...
+## schema goes here
+
+<!-- example plugin schema renderer below, ideally this will render on the left of the config examples 
+
+{% schema %}
+type: plugin
+data:
+  name: rate-limiting
+{% endschema %}
+-->
```
cloudjumpercat, line 42:
```suggestion
<!-- example plugin schema renderer below, ideally this will render on the left of the config examples -->
```

cloudjumpercat:
@fabianrbz I reset the branch so it would pull in the plugin changes you made to main. I added the schema placeholder, do you want to add that to this PR or should we merge this in first?

I don't know why, but adding a reference.md page to the rate limiting plugin folder caused the preview to display weirdly: https://deploy-preview-20--kongdeveloper.netlify.app/plugins/rate-limiting/ I tried hardcoding the index and reference links in the plugin layout page, but that didn't fix it locally.

fabianrbz:
I've added the changes as part of https://github.com/Kong/developer.konghq.com/commit/145b338114d0d82c0fc4fe5177844fe4fd41aa79

PR #21, opened 2024-07-17
Kong entities

lena-larionova:
Adding entity pages for:
* Services
* Routes
* Plugins
* Upstreams
* Targets - ~not sure about this one. I think that targets need upstreams to work - should targets be covered under upstreams instead?~ After some thinking, and our conversations on individual pages for intent, this is definitely a page - doesn't matter if it's short.
  * After the preview loaded, I learned that new topics don't get generated during a running build, I have to stop, `make clean`, and restart the build first.

https://deploy-preview-21--kongdeveloper.netlify.app/kong-entities/

Outstanding questions:
* How much info is too much? Where do we draw the line on what's info about an entity, and what's something else? For example, the plugins page goes into a lot of detail on many different aspects of plugins. Should all of that info live on different pages, eg:
  * Plugin entity
  * Plugin precedence
  * Plugin scopes
  * Custom plugins
* When there's a lot of content, the entity block gets pushed to the bottom of the page. Should we try to keep the content shorter so that it doesn't get lost? Or allow authors to specify the position of the block?
* Seeing some of these longer pages makes me want some kind of table of contents. Wonder if we could generate a collapsible ToC at the top of the page somewhere, and enable it with something like `toc: true`?
*


lena-larionova:
Just tried it again both ways:
* Build is already running with make run
* Created a certificates.md page, it doesn't appear in the preview
* Stopped the build, ran it again with make run, still no certificates page

PR #23, opened 2024-07-18
Entities landing page experiment

lmilan:


app/_kong_entities/service.md
```diff
@@ -5,14 +5,16 @@ entities:
 
 content_type: reference
 
+description: In Kong Gateway, a service is an abstraction of an existing upstream application. Services can store collections of objects like plugin configurations, and policies, and they can be associated with routes.
```
lena-larionova, line 8: I like this a lot. We can then require page to have a description (maybe limit it to some character count)? Then we can pull it in any place that we do link previews, including SEO reasons - this is the description that should get pushed to Google.

app/_includes/landing_pages/entity_card.md
```diff
@@ -0,0 +1,14 @@
+{% assign entity_id = include.config.entity %}
+
+{% assign entities = site.collections | first %}
+{% assign entity = entities.docs | where_exp: "doc", "doc.entities contains entity_id" | first %}
```
fabianrbz, line 4: this is great!
you can actually call collections by their name, e.g. `site.kong_entities`.

fabianrbz: weird, we are using it [here](https://github.com/Kong/developer.konghq.com/blob/main/app/kong-entities.html#L7).
which part isn't working for you? the `where_exp` or fetching the collection?

lmilan: I tried again and it works now, thanks! I think originally I may have only tried `site.collections.kong_entities` 😅

lena-larionova (approved):
lgtm, in case you're waiting on approval 😄

PR #27, opened 2024-07-22
Reference pages: routes and plugins; reference layout

lena-larionova:
Split out routes and plugins info into separate reference pages; create a placeholder reference layout (generic reference layout? vs a specific "entity reference" or something.)


fabianrbz:
@lena-larionova I can update the `_kong_entities` collection url to be `gateway/entities/:name/`, should we also rename the folder to `app/_gateway_entities`?

app/gateway/entities/routes/expressions.md
```diff
@@ -0,0 +1,162 @@
+---
```
fabianrbz, line 1: based on this url, would you expect `/gateway/entities/routes/` to be the current route entity reference page?

fabianrbz:
Should these pages live under `/gateway/entities/routes/` ?
* `/gateway/entities/routes/expressions/`
* `/gateway/entities/routes/routing/`

They are related to the kong entity route, but I don't know if they should be nested that deep, what about
* `/gateway/routing/`
* `/gateway/routing/expressions/`

Unrelated to the above, but would you consider these pages to be "reference pages" or "concept pages"?

PR #32, opened 2024-07-29
Add authentication landing page

lmilan:


app/_landing_pages/authentication.yaml
```diff
@@ -0,0 +1,106 @@
...
+                - type: text
+                  text: |
+                    Authentication is the process of verifying that a requester has permissions to access a resource. 
+                    As its name implies, API gateway authentication authenticates the flow of data to and from your upstream services. 
+                    
+                    Kong Gateway has a library of plugins that support the most widely used methods of API gateway authentication.
+                    
+                    Common authentication methods include:
+
+                    - Key Authentication
```
fabianrbz, line 30: I think it would be good to expand a bit more about authentication methods, this is the page where they are introduced to `Authentication` and we guide them towards different solutions.

We could provide the user a brief intro to the methods `SSO`, `OICD`, `LDAP`, `Basic Auth`, `Key Auth` etc. and even have specific pages for each of those where they can learn more about them and be linked to more specific tutorials, etc

lmilan: By specific page, so you mean a plugin overview page for each authentication plugin? Or something else? And do we want to do these for the MVP?

fabianrbz: I wasn't thinking about the plugins, but more about the methods. In particular SSO, OICD, LDAP, SAML.

PR #34, opened 2024-07-29
Generate tutorial list for landing pages

lmilan:


fabianrbz:
@lmilan this is great!
We probably want to do this in ruby instead of in the templates because it can get quite complicated. I'll assign this to me and we can work together on it.

fabianrbz:
Closing this, I've added an implementation in ruby [here](https://github.com/Kong/developer.konghq.com/commit/6664037b8a341161f0469f42703ca1ff89ff9752)

PR #35, opened 2024-07-30
[Info Box] Add info_box tag with hardcoded information

fabianrbz:


mheap:
@fabianrbz Why do we need the tag here if all it does it render an include?

fabianrbz:
@mheap the plan is for the tag to create a drop, so we can gather all the information we need and pass it to the template, give me a minute.

fabianrbz:
@mheap just pushed an example, the most relevant bit is [this](https://github.com/Kong/developer.konghq.com/pull/35/files#diff-e13d36c83a500e1f40e9c04a1c6ffde02ea1529e3dcb0c5c2cff39d2a1a85212R50-R57)

We should get rid of the data files, except for `tags.yml`.
I believe all the metadata should live on the pages, however right now there's no way of accessing `tools` or `products`.
We can't use Jekyll collections for those because they are yaml files and Jekyll collections only work with markdown files.

**What we need**
Is to create specific folders for them and use generators  to render the pages, e.g.
```
app/
  _products/
  _tools/
```
Then we can build specific generators for them, build the pages and keep an array for each on the site, i.e. `@site.data.products << page` in the same way we generate the plugin hub in the docs.

We'll need to do the same for `entities` and `plugins` too at some point, because those pages will be versioned. Plugins also have a another issue and that is that we have multiple files for them and Jekyll Collections break in that case.

mheap:
In this case would the Gateway landing page actually be in `_products` and the deck one be in `_tools`?

PR #48, opened 2024-08-08
Rate Limiting landing page

cloudjumpercat:
- Revised the existing RL landing page (demo)
- I wanted to add general info about what RL is, what it does, and then some info about which solution a user should choose

### DIFF
--- a/app/_landing_pages/demo.yaml
+++ b/app/_landing_pages/demo.yaml
@@ -19,30 +19,32 @@ content:
                 text: "What is Rate Limiting?"
               blocks:
                 - type: text
-                  text: Rate Limiting is a technique for preventing individual
-                    users or clients from overloading a system by making
-                    too many requests over a specific time period.
+                  text: Rate limiting controls the number of API requests a client can make within a specified time frame. 
                     It's a critical function that helps maintain
                     the performance and availability of web
                     services and APIs by controlling the request traffic.
       - column:
           - type: mermaid
             config:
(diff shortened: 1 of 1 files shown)


cloudjumpercat:
@fabianrbz This is failing with the following error: "Liquid Exception: Error rendering {% plugin %} on page: /opt/build/repo/app/_landing_pages/authentication.yaml. The plugin `` doesn't exist. in /opt/build/repo/app/_layouts/landing_page.html"

I tried seeing if I could find the problem on the authentication landing page, but I couldn't figure out what it was. Do you know how I can fix this so I can see the preview link?

lena-larionova:
> @fabianrbz This is failing with the following error: "Liquid Exception: Error rendering {% plugin %} on page: /opt/build/repo/app/_landing_pages/authentication.yaml. The plugin `` doesn't exist. in /opt/build/repo/app/_layouts/landing_page.html"
> 
> I tried seeing if I could find the problem on the authentication landing page, but I couldn't figure out what it was. Do you know how I can fix this so I can see the preview link?

It's a formatting issue, should be `slug` instead of `name`. The other problem is that it expects these plugins to exist already, which they don't. 

```
  - row:
      - column:
          - type: plugin
            config:
              slug: rate-limiting
```

lena-larionova:
Fixed here: https://github.com/Kong/developer.konghq.com/pull/50

You should be able to rebase to pick it up now.

PR #49, opened 2024-08-09
Rate limiting advanced plugin

lena-larionova:
Lots of questions/comments:

* Right now we don't have support for multiple pages within a plugin directory, the pages conflict with each other:
    ```
     The following destination is shared by multiple files.
    22:46:43 jekyll.1 |                     The written file may end up with unexpected contents.
    22:46:43 jekyll.1 |                     /github-repos/developer.konghq.com/dist/plugins/rate-limiting-advanced/index.html
    22:46:43 jekyll.1 |                      - /github-repos/developer.konghq.com/app/_kong_plugins/rate-limiting-advanced/index.md
    22:46:43 jekyll.1 |                     - github-repos/developer.konghq.com/app/_kong_plugins/rate-limiting-advanced/reference.md
    ```
    This is the same problem that Diana ran into here: https://github.com/Kong/developer.konghq.com/pull/20
    
    I tried messing around with the jekyll config file and the collections but couldn't figure it out. Haven't worked with collections before so that is a bit of a mystery - can collections be modified to use pieces of content here and there? For example, right now the plugins index is pulling in two entries for the RLA plugin, its index page and its reference page, but the links are both to the landing page, and the content is from reference (all very confusing).

* Rate Limiting Advanced landing page: so much content! This feels like too much detail to get into on a landing page, but at the same time, I don't want it to get lost. Should all of this live on different pages? If we're not nesting anything other than index, reference, and maybe API doc under each plugin, where would it all go? 

* Added metadata for `tier: enterprise` and `konnect_compatible: true`. Could also make it `konnect: true`. Since the pricing model is changing, we don't need to worry about Konnect Free/Plus/Enterprise tags here and can just have a true/false flag.
* How do we want to portray multiple examples? Right now, on the reference page, they're listed out in a row. Should there be a dropdown to select an example, or is that too many dropdowns? 
* Non-plugin specific: should every single page have a `description` key with a short description? I'd say yes, for search metadata reasons.


lena-larionova:
Implemented what we talked about on Friday: split content, moved examples into metadata. 

When splitting the content, went down a rabbit hole of need a bunch more plugin pages, so now we have stubs for all the Rate Limiting plugins - that seems reasonable.

Questions:
* Should we display all plugins as cards on reference pages, tutorials, etc as well? For consistency? Or should they appear in related links?
* Should we have a separate `plugin_reference` layout, or should we use `plugin`?
* I turned the plugin's reference.md file into yaml. I can't think of why not - most of that content should be generated, so yaml seems to make sense.

@fabianrbz feel free to make changes here directly, if you're fixing anyway, and merge this whenever to get them in sooner.

fabianrbz:
> Implemented what we talked about on Friday: split content, moved examples into metadata.

> When splitting the content, went down a rabbit hole of need a bunch more plugin pages, so now we have stubs for all the Rate Limiting plugins - that seems reasonable.
That's great, thanks! 

Questions:

> Should we display all plugins as cards on reference pages, tutorials, etc as well? For consistency? Or should they appear in related links?

This is something we can discuss with the rest of the team. I don't think that cards are needed because it might give them too much visibility in reference pages and tutorials.

> Should we have a separate plugin_reference layout, or should we use plugin?

I added a two, although the one for the reference doesn't have much, we'll see if we need it in the future.

>I turned the plugin's reference.md file into yaml. I can't think of why not - most of that content should be generated, so yaml seems to make sense.

That makes sense.

> @fabianrbz if it helps at all - feel free to make changes here directly, if you're fixing plugin collections/references anyway, and merge this whenever to get them in sooner.
I've made some changes, will merge this soon.

PR #54, opened 2024-08-13
Tools landing page

cloudjumpercat:
Created a landing page for Kong tools

### DIFF
--- a/app/_landing_pages/tools.yaml
+++ b/app/_landing_pages/tools.yaml
@@ -1,0 +1,135 @@
+metadata:
+  title: "Kong Tools"
+  tags:
+    - deck
+    - kic
+    - terraform
+    - konnect-api
+    - admin-api
+content:
+  - row:
+      - column:
+          - type: header
+            config:
+              type: h1
+              text: "Kong Tools"
+              align: center
+
+  - row:
+      - column:
+          - type: structured_text
+            config:
+              header:
+                text: "What are Kong tools?"
+              blocks:
+                - type: text
+                  text: Rate limiting controls the number of API requests a client can make within a specified time frame. 
+                    It's a critical function that helps maintain
+                    the performance and availability of web
+                    services and APIs by controlling the request traffic.
+      - column:
+          - type: mermaid
+            config:
+              alt_text: "A diagram that shows how various Kong tools work, with Konnect as an example"
+              diagram: |
+                graph TD
+                subgraph "Deployment & Configuration"
+                    A[Terraform] -->|Configures| B(Konnect)
+                end
+
+                subgraph "Management"
+                    B -->|Managed by| C(KIC)
+                    B -->|Configures Gateway entities| D(Admin API)
+                end
+
+                subgraph "Configuration Updates"
+                    F(decK) -->|Updates| B
+                    B -->|Updated by| E(Konnect API)
+                end
+
+  - row:
+      - column:
+          - type: structured_text
+            config:
+              header:
+                text: "Which Kong tools should I use?"
+              blocks:
+                - type: text
+                  text: "You can use multiple tools, depending on your use case:"
+                - type: unordered_list
+                  items:
+                    - "**Admin API:** Allows you to configure and manage Kong Gateway entities."
+                    - "**Konnect API:** Allows you to configure and manage Konnect-specific components, like Dev Portal or API Products."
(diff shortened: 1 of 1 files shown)


cloudjumpercat:
@lmilan Here's the tools landing page! Feel free to add or edit anything you'd like! I couldn't think of a lot of words to go on this page.

app/_landing_pages/tools.yaml
```diff
@@ -0,0 +1,135 @@
...
+                - type: text
+                  text: Rate limiting controls the number of API requests a client can make within a specified time frame. 
+                    It's a critical function that helps maintain
+                    the performance and availability of web
+                    services and APIs by controlling the request traffic.
+      - column:
+          - type: mermaid
+            config:
+              alt_text: "A diagram that shows how various Kong tools work, with Konnect as an example"
+              diagram: |
```
mheap, line 34: What's the difference between Konnect and Konnect API?

app/_landing_pages/tools.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - column:
+      - type: card
+        config:
+          title: Terraform
+          icon:  https://www.svgrepo.com/show/376353/terraform.svg
+          cta: https://docs.konghq.com/konnect/reference/terraform/
+    - column:
+      - type: card
+        config:
+          title: KIC
```
mheap, line 103: This is an interesting one. KIC is a tool, but it's also a product. I'd drop KIC for now

app/_landing_pages/tools.yaml
```diff
@@ -0,0 +1,135 @@
...
+      
+  - row:
+      - column:
+          - type: header
+            config:
+              type: h2
+              text: Tutorials
+          - type: tutorial_list
+            config:
+              - tag: deck
```
fabianrbz, line 115: This won't work with the current implementation, but we can change that :)
```
config:
  - tag: # filters by tag, e.g. rate-limiting, etc. it only supports one, omit it for all tags
  - product: # filters by product, it only supports one , omit for all products
```

we can update to also support: `tool: # filter by tool, omit for all tools`

We could also make them accept more than one value, but we should use the plural for the key then

app/_landing_pages/tools.yaml
lena-larionova: I'm guessing this is referring to the UI, but I'd like clarity on that too.

app/_landing_pages/tools.yaml
```diff
@@ -0,0 +1,150 @@
...
+              blocks:
+                - type: text
+                  text: "You can use multiple tools, depending on your use case:"
+                - type: unordered_list
+                  items:
+                    - "**Admin API:** Allows you to configure and manage Kong Gateway entities."
+                    - "**Konnect API:** Allows you to configure and manage Konnect-specific components, like Dev Portal or API Products."
+                    - "**decK:** Allows you to visualize configurations and apply changes in a declarative manner to Kong Gateway or Konnect."
+                    - "**Terraform:** Allows you to manage infrastructure as code and automated deployments to streamline setup and configuration of Konnect and Kong Gateway."
+                    - "**KIC:** Allows you to manage ingress traffic and routing rules."
```
lena-larionova, line 65: I think we talked about not including lists like this (entity landing page was the example?), since all of this information should be on the cards themselves.

app/_landing_pages/tools.yaml
```diff
@@ -0,0 +1,150 @@
...
+          - type: structured_text
+            config:
+              header:
+                text: "What are Kong tools?"
+              blocks:
+                - type: text
+                  text: Rate limiting controls the number of API requests a client can make within a specified time frame. 
+                    It's a critical function that helps maintain
+                    the performance and availability of web
+                    services and APIs by controlling the request traffic.
```
Guaris, line 29:
```suggestion
                  text: There are three types of tools availble for interacting with Kong Gateway
                    APIs: The Admin API and the control plane configuration API. 
                    SDKs: 
                    CLIs: CLI utilities like decK are available for configuring KGW and Konnect
```

Really rough idea of a way we could split this into being about the three tool types if we decide to go that route. Then we can split the cards at the bottom by those types.

app/_landing_pages/tools.yaml
fabianrbz: I ended up implementing the changes I mentioned above ^

lena-larionova:
@cloudjumpercat can you rebase and update this page + diagram based on our latest styles?

PR #56, opened 2024-08-15
fix: Rate Limiting Tiers tutorial

cloudjumpercat:
I went back and changed the following things about the rate limiting tiers tutorial:

- The prereqs listed a service, which isn't required for this tutorial. I added a `consumer` and `consumer-group` prereq instead, but I think I formatted it incorrectly, because it wasn't running for me locally. (cc @fabianrbz )
- I added `deck gateway sync` to the prereq so they sync that file before doing any of the work in the tutorial body.
- I also added `deck gateway sync` at the end of the tutorial


app/_data/entity_examples/consumer-groups/example-consumer-group.yaml
lena-larionova, line 1: Looks like the build expects the file to be `.yml` instead of `.yaml`

fabianrbz: @cloudjumpercat that's the problem, I renamed the file accordingly

PR #58, opened 2024-08-16
Chore: Cleaning up tutorials

lena-larionova:
Fixing a broken tutorial, switching the OIDC tutorial (which doesn't exist on our docs site anymore) with the SSO content that replaced it, and various minor cleanup for tutorials.

The SSO tutorial has prereqs that are unique to it. I added them in as manually for now, didn't want to fail the build. How do we want to handle those?


fabianrbz:
do we want to split this into two tutorials? one for OIDC and one for SAML?

PR #60, opened 2024-08-16
fix: Add tiers to RL tiers tutorial

cloudjumpercat:
- make more tiers and add them to the steps instead of prereqs
- expand Q&A to include use case
- more verbose about what is being configured in the steps

What I'd like:

- I'd like the entity examples to be cumulative, and I don't think that's supported. Ex. In the first step, I add a consumer. Then I add the consumer to the group. I want that second step to contain both the consumer AND the consumer group in the file example. So at the end of the tutorial, you have the FULL file contents. 
- I'd like to configure multiple consumers/consumer groups/etc. at a time in an entity example, and I don't think I can currently do that (it only displays the last one in the rendered output).


cloudjumpercat:
I was working on adding [tiers to the RL tiers tutorial](https://github.com/Kong/developer.konghq.com/pull/60) and it was difficult. Here's what I wished:
When each step is bit by bit (ex. add one consumer Tom then add the next consumer Jan in the next step) I can't just copy/paste each entity example into my kong.yaml file because at some point, it throws an error because it wants the values to be nested.

For example, here's what I would copy and paste:
```
_format_version: '3.0'
consumers:
  - username: Tom
consumers:
  - username: Jan
```

decK only picks up the last one, Jan and doesn't make Tom. I'd need to nest the formatting like this to make it work:
```
_format_version: '3.0'
consumers:
  - username: Tom
  - username: Jan
```

I don't intuitively understand this from reading the tutorial though since I'm not a decK power user. I found this [hidden doc](https://github.com/Kong/go-apiops/blob/main/deckformat/fileformat.md#metadata) to be the most helpful with understanding how to nest and format the decK file.

**What I'd prefer:**
Each step builds and adds to the example file that was in the previous step, so you can copy/paste the whole thing and see how it nests as you go through the tutorial.
For example:
1. Create consumer Tom
```
 _format_version: '3.0'
consumers:
  - username: Tom
```
1. Create consumer Jan
```
_format_version: '3.0'
consumers:
  - username: Tom
  - username: Jan
```
1. Assign them to the team consumer group
```
_format_version: '3.0'
consumer_groups:
- name: team
consumers:
  - username: Tom
    groups:
    - name: team
  - username: Jan
    groups:
    - name: team
```
1. Enable the plugin on the consumer group
```
_format_version: '3.0'
consumer_groups:
- name: team
  plugins:
  - config:
      limit:
      - 1000
      retry_after_jitter_max: 0
      window_size:
      - 3600
      window_type: sliding
    name: rate-limiting-advanced
consumers:
  - username: Tom
    groups:
    - name: team
  - username: Jan
    groups:
    - name: team
```

mheap:
In this case, why would you create consumers, then consumer groups, then update consumers?

How about:

1. Create Consumer groups for each tier:
```
_format_version: '3.0'
consumer_groups:
- name: team
```

2. Create consumers that are on that team:

```
consumers:
  - username: Tom
    groups:
    - name: team
  - username: Jan
    groups:
    - name: team
```

3 Enable the plugin on the consumer group
```
plugins:
  - consumer_group: test
     config:
      limit:
      - 1000
      retry_after_jitter_max: 0
      window_size:
      - 3600
      window_type: sliding
    name: rate-limiting-advanced
```

You can append each of these to the end of a file

cloudjumpercat:
> In this case, why would you create consumers, then consumer groups, then update consumers?
> 
> How about:
> 
> 1. Create Consumer groups for each tier:
> 
> ```
> _format_version: '3.0'
> consumer_groups:
> - name: team
> ```
> 
> 2. Create consumers that are on that team:
> 
> ```
> consumers:
>   - username: Tom
>     groups:
>     - name: team
>   - username: Jan
>     groups:
>     - name: team
> ```
> 
> 3 Enable the plugin on the consumer group
> 
> ```
> plugins:
>   - consumer_group: test
>      config:
>       limit:
>       - 1000
>       retry_after_jitter_max: 0
>       window_size:
>       - 3600
>       window_type: sliding
>     name: rate-limiting-advanced
> ```
> 
> You can append each of these to the end of a file

@mheap I'll try that out!

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
```diff
@@ -38,57 +38,110 @@ tags:
...
-
-    This configuration sets the rate limit to five requests (`config.limit`) for every 30 seconds (`config.window_size`).
+   If everything looks right, synchronize them to update your Kong Gateway configuration:
+   ```sh
+   deck gateway sync deck_files
+   ```
+
+1. Enable the Rate Limiting and Rate Limiting Advanced plugins for each tier:
+  ```yaml
+  plugins:
```
cloudjumpercat, line 110: Something isn't working in the formatting here when it's rendered.

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
```diff
@@ -38,57 +38,110 @@ tags:
...
+      limit: 500
+      window_size: 30
+      window_type: sliding
+      retry_after_jitter_max: 0
+    name: rate-limiting-advanced
+  - name: rate-limiting
+    config:
+      second: 5
+      hour: 10000
+      policy: local
```
cloudjumpercat, line 136: Is this configured correctly? I feel like the rate limits for this one probably need to be adjusted to less than free, or do we even need this? Maybe global policy vs. policy for consumer groups is a different tutorial.

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
```diff
@@ -102,3 +155,20 @@ Consumers that are not in a consumer group default to the Rate Limiting advanced
    deck gateway sync deck_files
    ```
 
+1. Run the following command to test the rate limiting as the consumer:
+
+  ```sh
+  for _ in {1..6}; do curl -i http://localhost:8000/example_route -H 'apikey:example_key'; echo; done
+  ```
```
cloudjumpercat, line 162: I'm not sure what this does, is there a way I should modify it so that we can test each tier? And maybe a user who isn't in a tier?

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
fabianrbz: @cloudjumpercat I fixed the issues, and left an explanation in the commit message.
There's a caveat that I forgot to mention in the message though, and that is that if the content of a list item starts with a codeblock, there must be an empty line between the list item "title" and the code block. i.e.


```
1. Enable the Rate Limiting and Rate Limiting Advanced plugins for each tier:
    
   ```yaml # 3 spaces, aligned to Enable
          plugins:
   ```
```

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
fabianrbz: That depends on the purpose of the tutorial, if it's just about consumer groups, you don't need the rate-limiting plugin.

app/_tutorials/add-rate-limiting-tiers-with-kong-gateway.md
fabianrbz: this needs to be updated, I think this comes from the RL on a consumer tutorial. The idea there was to test that a specific consumer reached the quota and then received 429 afterwards.

It needs to be adjusted to this particular tutorial, you can try with at least one tier

fabianrbz (approved):

PR #64, opened 2024-08-21
Render schemas on gateway entities

fabianrbz:
Render the schema on entity pages using the new
spec-renderer and pulling the information from Konnect.

It works similarly to how the api specs work in the docs.
We keep a file `app/_data/konnect_oas_data.json` with the products'
information we have in Konnect so we can retrieve their API specs.

The file `app/_data/konnect_product_ids.yml` contains a mapping
of each API spec and its correpsonding product id in Konnect. This
provides a way to access and filter them easily.

The schema is rendered using a vue component that requires 3 arguments:
* konnect product id
* the version id of the specific product
* the path to the schema in the spec

The platform pulls this information from the different configuration
files based on the following metadata in a page's frontmatter:

```
schema:
  api: gateway/admin-ee
  path: /schemas/Consumer
```

The key `api` maps to the following entry in app/_data/konnect_product_ids.yml
```
gateway:
  admin-ee: 'konnect_product_id'
```

And the corresponding version information comes from
app/_data/konnect_oas_data.json

**NOTES:**
* This won't work in preview apps/production due to CORS, we need to ask Engineering to add an exception
* The vue component I'm using adds an extra `title`, `description` and `examples`. A  new prop will allow us to disable the examples, but we'll need to use a different component if we want to get rid of the `title` and `description`. Unfortunately, it's not being exported in the package. We could hide the title and example using css for the demo.
* We don't have a schema for Consumer Groups in Gateway EE
* You need to set `VITE_PORTAL_API_URL` (setting it in a .env file would do), its value should be prod's dev portal url

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+VITE_PORTAL_API_URL='https://developer.konghq.com/'
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ make install-prerequisites
 # Install dependencies
 make install
 
+# Create local .env file
+# OAS Pages require VITE_PORTAL_API_URL to be set in your current environment, it should match the Kong supplied portal URL
+cp .env.example .env
+
 # Build the site and watch for changes 
 make run
 ```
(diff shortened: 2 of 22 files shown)


fabianrbz:
Screenshot 2024-08-21 at 09 30 20

PR #71, opened 2024-08-27
decK landing page: further improvements + real text

lena-larionova:


fabianrbz:
@lena-larionova I've merged main into the branch and updated the page accordingly

fabianrbz:
Should we use one `faqs` block instead of multiple rows + faqs?

PR #79, opened 2024-09-13
decK guides

lena-larionova:
Porting some decK guides over to fill the how-to guides filter on the decK landing page.

Previews:
[decK](https://deploy-preview-79--kongdeveloper.netlify.app/deck/)
[Use environment variables with decK](https://deploy-preview-79--kongdeveloper.netlify.app/how-to/deck-environment-variables/)
[Customise object defaults in decK](https://deploy-preview-79--kongdeveloper.netlify.app/how-to/deck-object-defaults/)
[Manage decK with Docker](https://deploy-preview-79--kongdeveloper.netlify.app/how-to/manage-deck-with-docker/)

Lots of questions/discussion points:

* 2 out of 3 of these pages felt more like references than guides. They don't really fit in our current "how-to" paradigm, as there's a mix reference/explanation/examples.
  * for example - object defaults is probably actually two pages: How to set custom object defaults + Reference on decK object default handling
  * Ran into some small issues with the decK prereq: 
    * If I set the tool value as "deck", the "install deck" prereq comes up. This isn't relevant for the "Manage decK with Docker page", which is explicitly referencing a different way of installing decK. However - this page might actually be a reference and then this is a moot point.
    * The line at the end of the prereq ("decK enables a simpler configuration and troubleshooting process allowing you to focus on the tutorial and not the tools") is redundant for a decK guide.

* How long/short can the answer be in the tldr section? For an extreme example, just to see it break, I put a bunch of paragraphs into the object defaults guide. That seems like too much. But I also don't like that if I break it up, the longer explanation appears after the Prerequisites section.

* The paragraph spacing on the long explanation text (text that appears right after Prerequisites) is really big. You can see it here: https://deploy-preview-79--kongdeveloper.netlify.app/how-to/deck-environment-variables/


app/_how-tos/custom-deck-object-defaults.md
```diff
@@ -0,0 +1,153 @@
...
+{{site.base_gateway}} and with {{site.konnect_saas}}.
+
+{:.important}
+> **Important:** This feature has the following limitations:
+* Custom plugin object defaults are not supported.
+* If an existing property's default value changes in a future {{site.base_gateway}} release,
+decK has no way of knowing that this change has occurred, as its `defaults`
+configuration would overwrite the value in your environment.
+
+1. In your `kong.yaml` configuration file, add an `_info` section with
```
fabianrbz, line 53: @lena-larionova we changed this to be h2s instead of a list, each h2 (and the content after it) is treated as a step.

app/_how-tos/custom-deck-object-defaults.md
```diff
@@ -0,0 +1,153 @@
+---
+title: decK object defaults reference
+
+works_on:
+    - on-prem
+    - konnect
+
+tags:
```
fabianrbz, line 8: we decided to set the `product` as `api-ops` too for deck's how-tos, that way we could filter how-tos etc by `product` and `tool`.

app/deck/environment-variables.md
```diff
@@ -0,0 +1,69 @@
...
+
+content_type: reference
+layout: reference
+
+works_on:
+    - on-prem
+    - konnect
+
+tags:
+  - api-ops
```
fabianrbz, line 12: same on reference pages, product should be `api-ops` too.

fabianrbz:
> The line at the end of the prereq ("decK enables a simpler configuration and troubleshooting process allowing you to focus on the tutorial and not the tools") is redundant for a decK guide.

Should we drop it entirely? Or render it when the product/tool combination is not `product: -api-ops`, `tools: deck`


> The paragraph spacing on the long explanation text (text that appears right after Prerequisites) is really big. 

I'll fix this

app/deck/distributed-config.md
```diff
@@ -0,0 +1,115 @@
...
+For example:
+
+| Use case | Description |
+|----------|-------------|
+| Single Kong installation whose configuration is built by multiple teams | For example, one team manages the inventory API, while another team manages the users API. Both APIs are exposed via {{site.base_gateway}}. <br> Team Inventory needs to manage its configuration without worrying about the configuration of the other team, and vice versa. The two teams can push out configuration changes on their own schedules. |
+| Very large configuration | You have a large set of entities - hundreds, maybe thousands - and it's very difficult to manage them all in one file. You can use decK to break up the configuration into different files with specific purposes. |
+| Separation of consumer management | You want to declaratively manage all of the configuration for Kong except consumers and their credentials. This could be for any number of reasons: <br> - The consumers are being managed by another service <br> - The consumers are dynamically created <br> - The number of consumers is so large that it makes no sense to manage them in a declarative fashion | 
+
+## Methods for managing distrubuted configuration
```
lena-larionova, line 36:
```suggestion
## Methods for managing distributed configuration
```

PR #82, opened 2024-09-18
Insomnia tutorials

lmilan:
https://deploy-preview-82--kongdeveloper.netlify.app/how-to/generate-a-collection-from-an-api-specification/

### DIFF
--- a/app/_how-tos/generate-a-collection-from-an-api-specification.md
+++ b/app/_how-tos/generate-a-collection-from-an-api-specification.md
@@ -10,13 +10,14 @@ tags:
 
 tldr: 
   q: How do I generate a collection from an API spec?
-  a: Create or import an API spec and define environment variables, then click cog icon and click Generate collection.
+  a: Create or import an API spec as a design document and define environment variables, then click settings icon and click <b>Generate collection</b>. Generating a collection from a document allows you to test your requests while still working on the design.
 
 prereqs:
   inline:
     - title: Create an API specification
       content: |
         You can either [create an empty document]() and design your specification from scratch, or [import an existing specification](). This example uses the [Konnect API Products specification](https://docs.konghq.com/konnect/api/api-products/latest/).
+      icon_url: /assets/icons/insomnia/design.svg
 
 ---
 
--- a/app/_includes/prereqs/products/insomnia.md
+++ b/app/_includes/prereqs/products/insomnia.md
@@ -1,8 +1,7 @@
 {% assign summary='Insomnia' %}
 {% capture details_content %}
-  This tutorial requires Insomnia. 
-  If you don't have it installed yet, see the [installation instructions]() for your OS.
+  [Download](https://insomnia.rest/download) and install Insomnia.
 {% endcapture %}
 
 
-{% include how-tos/prereq_cleanup_item.html summary=summary details_content=details_content %}
+{% include how-tos/prereq_cleanup_item.html summary=summary details_content=details_content icon_url='/assets/logos/brand/Insomnia.svg'%}


app/_how-tos/generate-a-collection-from-an-api-specification.md
```diff
@@ -0,0 +1,49 @@
...
+
+tldr: 
+  q: How do I generate a collection from an API spec?
+  a: Create or import an API spec and define environment variables, then click cog icon and click Generate collection.
+
+prereqs:
+  inline:
+    - title: Create an API specification
+      content: |
+        You can either [create an empty document]() and design your specification from scratch, or [import an existing specification](). This example uses the [Konnect API Products specification](https://docs.konghq.com/konnect/api/api-products/latest/).
```
cloudjumpercat, line 19: I think this could be a great opportunity to use [Kong Air](https://github.com/Kong/KongAir/blob/main/sales/bookings/package.json) for tutorials. We could use one of the specs in that repo for a prereq

app/_how-tos/generate-a-collection-from-an-api-specification.md
```diff
@@ -0,0 +1,49 @@
...
+	"base_url": "https://us.api.konghq.com/v2",
+	"bearerToken": "<my token>"
+}
+```
+
+## 2. Generate a collection
+
+Open your document and click the cog icon next to the **Preview** button. A collection of requests in created in the **COLLECTION** tab. 
+
+If the operations in the API spec have tags, the requests are organized into folders based on these tags. If an operation has multiple tags, the request is duplicated and appears in each folder. You can reorganize the requests as needed. For more information, see [Collections]().
```
cloudjumpercat, line 40:
```suggestion
Open your document and click the settings icon next to **Preview**. Click **Generate collection**. A collection of requests in created in the **COLLECTION** tab. 

If the operations in the API spec have tags, the requests are organized into folders based on these tags. If an operation has multiple tags, the request is duplicated and appears in each folder. You can reorganize the requests as needed. For more information, see [Collections]().
```

app/_how-tos/generate-a-collection-from-an-api-specification.md
cloudjumpercat: I wonder if we can provide the example spec contents in line so they can just copy/paste from this page. When I was doing this, I was a bit confused about where the spec went in Insomnia because I didn't realize I could copy/paste it right into the body, so I might say something along the lines of "Paste the spec directly into your document in Insomnia or import it in by clicking the **Import OpenAPI** drop down menu."

app/_how-tos/generate-a-collection-from-an-api-specification.md
```diff
@@ -0,0 +1,49 @@
...
+products:
+    - insomnia
+
+tags:
+  - collections
+  - documents
+
+tldr: 
+  q: How do I generate a collection from an API spec?
+  a: Create or import an API spec and define environment variables, then click cog icon and click Generate collection.
```
cloudjumpercat, line 13:
```suggestion
  a: Create or import an API spec and define environment variables, then click the Settings icon and click **Generate collection**. Generating a collection saves more time than creating one manually and you can ensure that all requests are consistent with the API specification, reducing errors.
```
Just wanted to add that suggestion because I think for me, I didn't understand the why behind generating vs just plain creating a collection.

app/_how-tos/generate-a-collection-from-an-api-specification.md
lmilan: I like the idea of using Kong Air, but when I import the specs in Insomnia I get a bunch of errors and warnings (which seem to be specific to Insomnia, because Swagger Editor doesn't report any errors on the same specs)

app/_how-tos/generate-a-collection-from-an-api-specification.md
lmilan: I think it's more document vs collection than generating vs creating a collection. Importing a spec as a collection would take about as much time as importing a spec as a document and then generating a collection. I think the main difference is that when you create a document you have access to design features, you can edit the spec and create unit tests, which you can't do from a standalone collection (if I understand correctly). So to me they're different use cases, one is design and the other is just sending requests. I'll update the how-to to make it clearer that we're talking about design documents.

PR #103, opened 2024-09-26
Insomnia: Chain requests + run collection

lmilan:


app/_includes/prereqs/collection.md
```diff
@@ -0,0 +1,8 @@
+1. [Create a new collection]().
+1. [Import]() the [KongAir Flights](https://raw.githubusercontent.com/Kong/KongAir/refs/heads/main/flight-data/flights/openapi.yaml) requests.
+1. Add the following content to the [base environment]():
+```json
+{
+    "base_url": "https://api.kong-air.com"
+}
+```
```
cloudjumpercat, line 8: @lmilan I revised this a bit for the write tests how to I was working on. I added the "Run in Insomnia" button for the spec so it's easier for them to add and also added a few more words about how to navigate to these things in the UI:

```
1. Import the KongAir Flights spec in Insomnia: 
    [![Run in Insomnia}](https://insomnia.rest/images/run.svg)](https://insomnia.rest/run/?label=&uri=https%3A%2F%2Fraw.githubusercontent.com%2FKong%2FKongAir%2Frefs%2Fheads%2Fmain%2Fflight-data%2Fflights%2Fopenapi.yaml)
1. Click the **Settings** icon for **SPEC** in the sidebar, and then click **Generate collection**. 
1. Click **Base environment** and then click the **Edit** icon.
1. Add the following content to the [base environment]():
```json
{
    "base_url": "https://api.kong-air.com"
}
```
```

lmilan: Two comments about this:
1. I'm wondering if we always need to create a document and then generate a collection, when all we need is a collection.
2. I didn't give too many details on purpose and gave links instead because these things are already documented in their own tutorials (or will be), so I thought we might want to avoid duplicating stuff.

cloudjumpercat: > I'm wondering if we always need to create a document and then generate a collection, when all we need is a collection.

That's a good question, and I feel like I don't understand Insomnia well enough to be able to answer this. As I'm working on the create a test how to, I've found the document/spec to be helpful to know what's required in the request. But that could just be me or the how to I'm working on. I don't want to block merging on any of these, so we can leave it how you have it now and I can add questions we have to the [Jira ticket](https://konghq.atlassian.net/browse/DOCU-4077).

> I didn't give too many details on purpose and gave links instead because these things are already documented in their own tutorials (or will be), so I thought we might want to avoid duplicating stuff.

I think I leaned more towards more details vs. links so the user can pretty much run down the whole page as a beginner with minimal detours to other pages. But I think this would be another great feedback question to ask the others once we have the rough draft of the Insomnia testing demo workflow on the dev site. I'll add this to the Jira ticket as well, so feel free to merge this PR whenever you'd like 🙂

lmilan: > That's a good question, and I feel like I don't understand Insomnia well enough to be able to answer this. 
My understanding is that a document is for designing the API, so for your how-to about creating tests, that makes sense, because the use case would be that you're checking the API spec against the implementation.
But if I'm just using Insomnia to try an API, without being involved in the design, then I would use a collection, because it's simpler, and I don't necessarily need to look at the spec.

> I've found the document/spec to be helpful to know what's required in the request.
You should have that information in the collection too, I have a spec that I imported as a collection, and the request bodies are populated with the examples in the spec, so I can just edit that.

PR #108, opened 2024-09-26
Create tests in Insomnia

cloudjumpercat:
- how to page covering how to create tests in Insomnia


app/_how-tos/write-data-type-tests.md
```diff
@@ -0,0 +1,74 @@
...
+```javascript
+const response1 = await insomnia.send();
+const body = JSON.parse(response1.data);
+expect(body).to.be.an('object');
+expect(body).to.have.property('meal_options'); 
+expect(body.meal_options).to.be.an('array');
+if (body.meal_options.length > 0) {
+    expect(body.meal_options[0]).to.be.a('string');
+}
+```
```
cloudjumpercat, line 73: Note for reviewers: This was created with ChatGPT. Although the test passes, I don't fully trust it or understand it. Is `meal_options` really nested? Is this really testing the right thing?

PR #113, opened 2024-10-03
Insomnia: Unit testing page

cloudjumpercat:
Page is intended to be a concept/reference hybrid, similar to the Consumers entity page.

I can't seem to get the concept bit to function correctly.


app/insomnia/unit-testing.md
```diff
@@ -0,0 +1,90 @@
...
+    API->>FS: Call GetFlightDetails("KA0284")
+    FS->>DB: Query for flight details
+    DB-->>FS: Return flight details
+    FS-->>API: Return flight details
+
+    alt Flight doesn't exist
+        DB-->>FS: Return error (404 not found)
+        FS-->>API: Return error response (404 not found)
+    end
+{% endmermaid %}
```
cloudjumpercat, line 53: ChatGPT helped me make this, but it went through many revisions. I think I'm unsure if this particular diagram is helpful to understanding unit testing. And I read external systems shouldn't be in unit testing, but I think some unit tests (like flight service) are using a database of sorts. So I was confused about that.

Guaris:
Based on conversation with Fellipe lets close this.

PR #114, opened 2024-10-04
Inso CLI landing page

lmilan:


app/_landing_pages/inso-cli.yaml
```diff
@@ -0,0 +1,119 @@
...
+        blocks:
+        - type: text
+          config: |
+            When you run a command, Inso CLI will look for Insomnia data in two places:
+            * In the .insomnia directory in the working directory. This directory is created automatically when [Git sync]() is set up.
+            * In the Insomnia app data directory if Insomnia is installed.
+
+            You can also use the `--workingDir` or `--src` options to specify a different data source. For more information, see the [Inso CLI reference]().
+
+  - header:
```
cloudjumpercat, line 45: Could you add an example with the rest of the commands from [this table](https://docs.insomnia.rest/inso-cli/cli-command-reference#command-overview)? Just so we have all the main commands on this page.

app/_landing_pages/inso-cli.yaml
```diff
@@ -0,0 +1,119 @@
...
+      text: "Frequently Asked Questions"
+      type: h2
+    columns:
+      - blocks:
+        - type: faqs
+          config:
+            - q: 
+              a:
```
cloudjumpercat, line 104: I can't think of any FAQ people might have about the CLI at the moment, maybe we could remove this for now?

lmilan: I'll remove it for now and we can revisit the cards when we have the reference pages

PR #115, opened 2024-10-07
Insomnia: Documents landing page

cloudjumpercat:
Why I formatted it this way:

- Documents really is a combo of different features (design, debug, test), so I wanted to highlight the capabilities on the outset and then dive into each capability as it's own section.
- I decided to link to the feature landing pages because I figured that the majority of the info belonged on those pages and these sections in the Documents page are just to get people oriented/learn what a document is, not a deep dive on what a test is or a spec. 
- For each feature section, I added how-to card links specific to that feature instead of using the how-to list at the end. 
- This page almost is a combo of a landing page and a "entity" type concept page.

What I'd like feedback on specifically:

- Is there too much content on the page? 
- Do you like how it's broken into sections by feature? Or is there a different format we should use?
- The how-to card links, are they too much? Would you rather have the auto generated how-to list?


app/_landing_pages/insomnia/documents.yaml
```diff
@@ -0,0 +1,239 @@
...
+                  text: | 
+                    To debug your API spec, you can generate a collection of requests from it and debug those. 
+                    
+                    In the sidebar, you'll see a list of the requests. When you select one, you can send requests and then see the response in the right menu. The center panel will populate with the following:
+                - type: unordered_list
+                  items:
+                    - Request URL
+                    - Request type
+                    - Base URL variable (defined in the OpenAPI specification)
+                    - Request body (if applicable)
```
lmilan, line 110: I think having both request URL and base URL i might be confusing. I would say base URL, endpoints, and query parameters. And the list is missing request headers.

app/_landing_pages/insomnia/documents.yaml
```diff
@@ -0,0 +1,239 @@
...
+              align: end
+  - header:
+      text: "Frequently Asked Questions"
+      type: h2
+    columns:
+      - blocks:
+        - type: faqs
+          config:
+            - q: What's the difference between a collection and a design document?
+              a: You can't create, lint, and edit a spec from a collection. If you want to use version control sync, you should use a collection instead.
```
lmilan, line 221: I don't understand this:
> If you want to use version control sync, you should use a collection instead.

Because you can sync a design document with Git.

lmilan (commented):
It looks good, I would probably include something about mocks too (maybe in the debug section?).

> Is there too much content on the page?

There is a lot, but I would say that all of it is useful.

> Do you like how it's broken into sections by feature? Or is there a different format we should use?

I like it, it makes sense to me.

> The how-to card links, are they too much? Would you rather have the auto generated how-to list?

Maybe that would be a solution if we want to shorten the page a bit, but they don't bother me. I would maybe change the 'Get started' titles, because at first when I was scrolling down the page I thought it was a duplicated section since they had the same title. So maybe 'Get started with design', and so on?

app/_landing_pages/insomnia/documents.yaml
```diff
@@ -0,0 +1,239 @@
...
+              alt_text: Image of document tests in a test suite.
+      - blocks:
+          - type: structured_text
+            config:
+              blocks:
+                - type: text
+                  text: | 
+                    You can create tests for the requests populated from your specification in the Tests tab of a design document. You can use Insomnia's pre-configured test examples by editing a test and pressing **Control + Space**.
+                    
+                    You can manually run an individual test or an entire test suite from Insomnia. You can also automate tests with the Inso CLI.
```
Guaris, line 168:
```suggestion
                    
                    You can manually run an individual test or an entire test suite from Insomnia ,and you can automate tests with the Inso CLI.
```

Guaris: Would it make sense to link to Inso CLi here?

app/_landing_pages/insomnia/documents.yaml
Guaris: To me the difference is that a design document houses three interconnected pieces together. So like the design document is an object that can contain a spec, a collection, and tests. Then you can interact with those pieces using the features of Insomnia. 

A collection feels like it is just a collection of endpoints that can either be generated, imported, and just manually pasted in. 

I don't think there are any differences between the collection, and the collection within the design document as far as capabilities are concerned.

app/_landing_pages/insomnia/documents.yaml
```diff
@@ -0,0 +1,239 @@
...
+          config:
+            title: Chain requests in Insomnia
+            cta:
+              text: Learn more
+              url: /how-to/chain-requests/
+              align: end
+  
+  - header:
+      type: h2
+      text: "Test API specs"
```
Guaris, line 153:
```suggestion
      text: "Test APIs"
```

app/_landing_pages/insomnia/documents.yaml
```diff
@@ -0,0 +1,239 @@
...
+    columns:
+      - blocks:
+          - type: structured_text
+            config:
+              header:
+                text: "Design Documents in Insomnia"
+              blocks:
+                - type: text
+                  text: | 
+                    Design Documents (also called Documents) hold specifications, API requests, and tests. They also offer the capability to [sync with git](/).
```
Guaris, line 16:
```suggestion
                    Design Documents (also called Documents) hold API specifications, API requests, and tests. 
```

PR #120, opened 2024-10-16
Add breadcrumbs

fabianrbz:
Add breadcrumbs and structured data.

**Implementation details:**
Breadcrumbs are rendered based on the `breadcrumbs` key in the metadata, which should be an array of urls.
The text of each crumb is the title of the page matching the corresponding url.

**Notes:**

* Breadcrumbs don't include Home and current page
* The structured data does contain the current page as the last item
* I replaced the links to index pages in how-tos, plugins and gateway entities with breadcrumbs

### DIFF
--- a/app/gateway/rate-limiting/strategies.md
+++ b/app/gateway/rate-limiting/strategies.md
@@ -10,6 +10,10 @@ plugins:
   - rate-limiting-advanced
   - ai-rate-limiting-advanced
   - graphql-rate-limiting-advanced
+
+breadcrumbs:
+  - /gateway/
+  - /gateway/rate-limiting/
 ---
 
 All rate limiting plugins support some subset of the following strategies:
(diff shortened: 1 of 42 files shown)


lena-larionova:
Should we have a breadcrumb going back to the previous URL segment for landing pages? For example:
https://deploy-preview-120--kongdeveloper.netlify.app/insomnia/test/ - I'd expect this to have a breadcrumb that takes me back to the insomnia landing page.

fabianrbz:
> Should we have a breadcrumb going back to the previous URL segment for landing pages? For example: https://deploy-preview-120--kongdeveloper.netlify.app/insomnia/test/ - I'd expect this to have a breadcrumb that takes me back to the insomnia landing page.

yes, we definitely should! I didn't add those to the Insomnia docs because they are still being worked on, but I'll add them.

app/gateway/rate-limiting/strategies.md
```diff
@@ -10,6 +10,10 @@ plugins:
   - rate-limiting-advanced
   - ai-rate-limiting-advanced
   - graphql-rate-limiting-advanced
+
+breadcrumbs:
+  - /gateway/
+  - /gateway/rate-limiting/
```
fabianrbz, line 16: @lena-larionova I wasn't sure about these rate-limiting ones, WDYT? We can always re-visit them when we migrate the content

PR #125, opened 2024-10-22
Insomnia: small demo fixes

cloudjumpercat:
Wanted to make a few small fixes so the demo would at least flow fairly smoothly tomorrow. They include:
- Scratch pad was missing use cases on the storage options landing page
- Debug landing page had no content, so I added some
- a few small fixes, like broken links


cloudjumpercat:
this is just some small fixes and the bigger content adds (like the debug page) will be reviewed and revised anyways as part of the polish phase, so I'm going to merge this in for the sake of having this for the demo

PR #128, opened 2024-10-23
Insomnia: dynamic variables page

cloudjumpercat:
I'm not sure how I feel about this page. I based it off of [Postman's page](https://learning.postman.com/docs/tests-and-scripts/write-scripts/variables-list/).

Preview URL: https://deploy-preview-128--kongdeveloper.netlify.app/insomnia/dynamic-variables/

Thoughts:

- It's important to have _something_ for dynamic variables, so users know where they can use them, how to pull them up, some rough idea of what types are available and so they know they can migrate variables when they import stuff from Postman.
- I don't know that putting them all in a table is the best idea, like what I was starting to do and what Postman has. I can see maybe a user searching the page for something they need ex. "I need a variable for time" and then they can find the variable and have a better idea what to search when they pull up faker in the UI. But would most users really want that table? Is this just one of those cases where I'm documenting it because 1.) it's what we've done 2.) our competitor has done it in this way.
- Some variables are a bit different, so I can understand why the table description could be helpful, but this is more of a fringe case. Ex. for the country code, Insomnia uses the ISO 3166-1 code but Postman uses a different ISO standard, but you wouldn't know this because the name of the variable is the exact same. Does this matter?

### DIFF
--- a/app/insomnia/dynamic-variables.md
+++ b/app/insomnia/dynamic-variables.md
@@ -1,0 +1,90 @@
+---
+title: About dynamic variables 
+
+content_type: reference
+layout: reference
+
+products:
+    - insomnia
+
+related_resources:
+  - text: 
+    url: 
+
+tags:
+  - test-apis
+
+faqs:
+  - q: 
+    a: 
+---
+
+
+Insomnia has a feature that allows you to use pre-configured, random variables in your requests, request body, or environment. To do this, it uses the [Faker](https://www.npmjs.com/package/@faker-js/faker) library.
+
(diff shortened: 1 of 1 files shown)


Guaris:
This doc should only exist if it can be automated. The copy can be added to another doc abotu testing.

cloudjumpercat:
I'm going to recommend we add like two sentences about dynamic variables either to Insomnia Concepts or the Glossary: "Dynamic variables allow you to use pre-configured, random variables in your requests, request body, or environment. To use dynamic variables, start typing in a response, response body, or environment variable and a drop down menu will display with the different dynamic variables you can use."

So we'll have that in lieu of this page, so I'll close this PR.

PR #149, opened 2024-11-07
Add frontmatter validations

fabianrbz:
Schemas for each content_type are defined in `app/_data/schemas/frontmatter/`.
These are basic validations, the idea is to add more as we write docs or find new validations.

A tool that validates the frontmatter (except for plugin pages and autogenerated content for now) lives in `tools/frontmatter-validator` and a GH action runs it as a check.

**Considerations:**
* `title` and `content_type` are required for all pages
* `description` isn't required for how-tos for now, they have a `tldr`
* reference pages require  `products` or `tools`
* `how_tos` require `tldr`, `works_on` if `products == gateway`

**Note:**
* It outputs some warnings, I think the problem is that [ajv](https://ajv.js.org/)  - which is the tool that we use to load the json-schemas and validate the date - runs in strict mode and has some weird rules defined.
* There are currently 46 errors. All of them are because those pages don't have a `description`, I need help fixing those.


app/_data/schemas/frontmatter/concept.json
```diff
@@ -0,0 +1,4 @@
+{
```
cloudjumpercat, line 1: If we decide to require tiers, we might want to require them for concepts as well as how to. 

Ex. if we had a concept page about Advanced Analytics, that would be a paid tier.

app/_data/schemas/frontmatter/base.json
```diff
@@ -0,0 +1,73 @@
...
+    "anyOf": [
+      { "properties": { "content_type": { "const": "how_to" } } },
+      { "required": ["auto_generated"] }
+    ]
+  },
+  "then": {
+    "required": ["content_type", "title"]
+  },
+  "else": {
+    "required": ["content_type", "title", "description"]
```
cloudjumpercat, line 71: Does this mean that description is required for all pages with `content_type`?

fabianrbz: this means that all those properties (`["content_type", "title", "description"]`) are required

app/_data/schemas/frontmatter/how_to.json
```diff
@@ -0,0 +1,45 @@
...
+      "required": [
+        "q",
+        "a"
+      ]
+    },
+    "min_version": {
+      "type": "object"
+    }
+  },
+  "if": {
```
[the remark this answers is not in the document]
fabianrbz: I'm not sure, I'll make a note about this

PR #158, opened 2024-11-19
WIP: Meta changelogs

Guaris:


lena-larionova:
@Guaris I would leave the changelogs alone for now. We need to revisit the format of our changelogs - does the categorization and the nesting etc even make sense? Plus, there's a lot of content added to these files all the time, so it would be better to migrate once we've settled on an approach and can easily migrate new additions to the file.

PR #159, opened 2024-11-20
WIP: Add vale

fabianrbz:
Add Vale to the site.

## Notes

* Same styles we have in the docs site, they live in .github/styles/
* Treat yaml files (landing pages) as markdown files so we can check those too
* Added new [ignoreBlocks](https://github.com/Kong/developer.konghq.com/blob/add-vale-config/.vale.ini#L21-L24) so we don't add terms like github and oss to the Dictionary.
* By default Vale ignores Frontmatter, I added a tool (not intended for local use) that replaces the content of markdown files with their frontmatter, so then we can run vale on them. More on [this commit](https://github.com/Kong/developer.konghq.com/pull/159/commits/71ff62025332ae866a3550935fbdb716ecbe0230)
    * It uses reads keys from `.github/styles/docs/Frontmatter.txt` and
      replaces the entire line with empty strings, so that we don't get
      errors on those lines (or need to add exceptions to our dictionary).
    * Unfortunately, there's some duplication in `.github/styles/docs/Frontmatter.txt` and
      `.github/styles/landing_pages/Components.txt` that we can't avoid, due
      to Vale's limitations.
* There are two checks for Vale, one for the content of the files and one for the frontmatter. I've combined both checks into the same `job` in order to avoid extra setup, the downside is that if the first one fails, the second is skipped.


fabianrbz:
Screenshot 2024-11-21 at 12 44 03
Screenshot 2024-11-21 at 12 43 40

cloudjumpercat:
I read through the caveats you listed, those all sound good to me.

PR #260, opened 2025-01-15
Feat: Key & Key Set entities

### LINKED ISSUE #217: Keys and key sets entities
Sources:
https://docs.konghq.com/gateway/api/admin-ee/latest/#/Keys/list-key
https://docs.konghq.com/gateway/latest/reference/key-management/#main

lmilan:
Fixes #217 

### Preview Links

https://deploy-preview-260--kongdeveloper.netlify.app/gateway/entities/key/
https://deploy-preview-260--kongdeveloper.netlify.app/gateway/entities/key-set/

### Checklist 

- [x] Every page is page one
- [x] All pages contain metadata
- [x] Updated [sources.yaml](https://github.com/Kong/developer.konghq.com/blob/main/tools/track-docs-changes/config/sources.yml). For more info review [track docs changes](https://github.com/Kong/developer.konghq.com/tree/main/tools/track-docs-changes)
- [x] All autogenereted instructions render correctly (API, decK, Konnect). 
- [x] Style guide (capital gateway entities, placeholder URLs) implemented correctly.

### DIFF
--- a/app/_gateway_entities/key-set.md
+++ b/app/_gateway_entities/key-set.md
@@ -7,17 +7,12 @@ entities:
 description: A Key Set is a collection of {{site.base_gateway}} Keys.
 
 related_resources:
-  - text: Keyring entity
-    url: /gateway/entities/keyring/
   - text: Key entity
     url: /gateway/entities/key/
 
 tools:
   - admin-api
   - konnect-api
-  - kic
-  - deck
-  - terraform
 
 api_specs:
   - gateway/admin-ee
@@ -30,5 +25,20 @@ schema:
 
 ---
 
-@todo
-You can assign one or many keys to a JSON Web Key Set. This can be useful to logically group multiple keys to use for a specific application or service. Key Sets are also the preferred way to expose keys to plugins because they tell the plugin where to look for keys or have a scoping mechanism to restrict plugins to just some keys.
+## What is a Key Set?
+
+A Key Set is a collection of {{site.base_gateway}} Keys.
+
+You can assign one or many keys to a JSON Web Key Set. This can be useful to logically group multiple Keys to use for a specific application or service. Key Sets allow you to give a plugin access to a specific list of Keys.
+
+## Schema
+
+{% entity_schema %}
+
+## Set up a Key Set
+
+{% entity_example %}
+type: key
+data:
+  name: example-key-set
+{% endentity_example %}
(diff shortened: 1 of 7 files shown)


app/_gateway_entities/key-set.md
```diff
@@ -30,5 +25,20 @@ schema:
...
 
-@todo
-You can assign one or many keys to a JSON Web Key Set. This can be useful to logically group multiple keys to use for a specific application or service. Key Sets are also the preferred way to expose keys to plugins because they tell the plugin where to look for keys or have a scoping mechanism to restrict plugins to just some keys.
\ No newline at end of file
+## What is a Key Set?
+
+A Key Set is a collection of {{site.base_gateway}} [Keys](/gateway/entities/key).
+
+You can assign one or many Keys to a Key Set. This can be useful to logically group multiple Keys to use for a specific application or service. Key Sets allow you to give a plugin access to a specific list of Keys.
+
```
lena-larionova, line 33: Can you add the list of supported plugins from the original doc here as well? https://docs.konghq.com/gateway/latest/reference/key-management/#key-sets-endpoint 

You could probably add them inline here + in the page metadata as related links. From the source doc, I can't really tell if Keys should also have these links - probably?

app/_gateway_entities/key.md
```diff
@@ -20,16 +18,28 @@ api_specs:
...
 schema:
   api: gateway/admin-ee
   path: /schemas/Key
-
 ---
 
-@todo
+## What is a Key?
+
+A Key object holds a representation of asymmetric keys, either public or private. When Kong or a Kong plugin requires a specific public or private key to perform certain operations, it can use this entity.
```
lena-larionova, line 29:
```suggestion
A Key object holds a representation of asymmetric keys, either public or private. When {{site.base_gateway}} or a Kong plugin requires a specific public or private key to perform certain operations, it can use this entity.
```

app/_gateway_entities/key-set.md
lmilan: Ah yes I forgot I needed to ask about this, I think that list may be wrong or out of date, because I only found a key set config parameter on two of these.

For keys I _think_ there's no specific config in the plugins, they just have access to keys by default, but on certain plugins you can specify a key set to restrict access.

lmilan: After looking into it a bit more I think the original doc is about JSON Web Key Sets in general, but not the Key Set entity specifically. Some plugins, like the OIDC one, have their own endpoint to manage JWKS, but they're not necessarily compatible with the Key Set entity. From what I can tell, the only plugins that can use the Key Set entity are JWT Signer, ACME, and JWE Decrypt, so I'll add these.

app/_gateway_entities/key-set.md
```diff
@@ -30,5 +25,25 @@ schema:
...
+Currently, Key Sets can be used with the following plugins:
```
lena-larionova, line 34:
```suggestion
Key Sets can be used with the following plugins:
```
Let's remove "currently" since the pages are versioned, so it's always "current", technically.

lena-larionova (approved):

PR #266, opened 2025-01-16
chore: Plugin metadata

### LINKED ISSUE #244: Plugins metadata
* Plugin minimum versions
* Plugin topologies
  * DB-less explanations from old plugins don't need to be part of metadata, they're actually just notes for auth plugins.

lena-larionova:
Fixes #244 

Adding the `on_prem` and `konnect_deployments` categories to plugin `topologies`.

The rule should now be:
* All plugins must have a `topologies` key with at least one entry
* At least one of `on-prem` or `konnect_deployments` must exist, but it doesn't have to have both
* Some plugins have no `konnect_deployments` key because they're not supported in konnect at all

### Checklist 

- [x] Every page is page one
- [x] All pages contain metadata
- [x] Updated [sources.yaml](https://github.com/Kong/developer.konghq.com/blob/main/tools/track-docs-changes/config/sources.yml). For more info review [track docs changes](https://github.com/Kong/developer.konghq.com/tree/main/tools/track-docs-changes)
- [x] All autogenereted instructions render correctly (API, decK, Konnect). 
- [x] Style guide (capital gateway entities, placeholder URLs) implemented correctly.

### DIFF
--- a/.github/styles/frontmatter/Keys.txt
+++ b/.github/styles/frontmatter/Keys.txt
@@ -6,6 +6,7 @@ faqs
 icon_url
 include_content
 insomnia_link
+konnect_deployments
 min_version
 next_steps
 prereqs
--- a/.github/styles/frontmatter/Keys.txt
+++ b/.github/styles/frontmatter/Keys.txt
@@ -9,6 +9,7 @@ insomnia_link
 konnect_deployments
 min_version
 next_steps
+on_prem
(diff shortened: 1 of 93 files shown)


fabianrbz:
I would add hybrid for completeness (and to be explicit about it), a quick search/replace should fix this

Guaris (approved):

PR #319, opened 2025-01-24
Security update process doc

### LINKED ISSUE #274: Security Update Process Reference
https://docs.konghq.com/gateway/latest/production/security-update-process/

lena-larionova:
## Description

Fixes #274 

Questions + ideas + notes:
* This doc is coming from Kong Gateway but the text keeps referring to "Kong Gateway and other Kong products". Can we rephrase this and make it a blanket statement without specifically mentioning Kong Gateway, or do we need to consult with someone in eng to make that decision?
  * Leave as-is
* We should mark external links in some way, both in text and in info boxes. The "CVSS" link in this doc is external but there's no way to tell without clicking it.
  * Created issue: [External link icons](https://github.com/Kong/developer.konghq.com/issues/342)
* Does this info belong on the dev site? I think yes, because it's no longer just a docs site. On the docs site, I might've said no.
  * Yes, it belongs here

@cloudjumpercat - this would be a good doc to link from the security landing page.

## Preview Links
https://deploy-preview-319--kongdeveloper.netlify.app/vulnerabilities/

## Checklist 

- [x] Every page is page one.
- [ N/A ] Tested how-to docs. If not, note why here. 

### DIFF
--- a/app/_data/schemas/frontmatter/base.json
+++ b/app/_data/schemas/frontmatter/base.json
@@ -18,7 +18,7 @@
     },
     "content_type": {
       "type": "string",
-      "enum": ["landing_page", "how_to", "reference", "concept", "plugin", "plugin_example", "api"]
+      "enum": ["landing_page", "how_to", "reference", "concept", "plugin", "plugin_example", "api", "policy"]
     },
     "description": {
       "type": "string"
--- a/app/_data/schemas/frontmatter/policy.json
+++ b/app/_data/schemas/frontmatter/policy.json
@@ -1,0 +1,4 @@
+{
+    "$id": "schema:policy",
+    "allOf": [{ "$ref": "schema:base" }]
+  }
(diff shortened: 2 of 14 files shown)


lena-larionova:
@fabianrbz Two new concerns after switching to `reference` and moving the page to the root of `app`. Looks like we need to think about how we handle pages that are applicable to all of our products:
* `content_type: reference` requires `products` and `tools` to be defined, but neither one is applicable here. I get front matter errors if I leave either of those keys out
* Do we want to have content pages just sitting directly in the `app` directory? If there are more than a few, then it could get messy.

fabianrbz:
> * content_type: reference requires products and tools to be defined, but neither one is applicable here. I get front matter errors if I leave either of those keys out
We could either remove the validation or set a specific key to bypass it.

> * Do we want to have content pages just sitting directly in the app directory? If there are more than a few, then it could get messy.
This one is up to us, we could also put them in a folder and set the `permalink` in the frontmatter.

fabianrbz (approved):

PR #352, opened 2025-01-28
Use consistent terminology for describing service applications

lena-larionova:
## Description

Adjusting the language/terminology around upstream/service/application, which all refer to the same thing.
Issue reported on Slack.

## Checklist 

- [x] Every page is page one.


lena-larionova:
Broken links not part of this PR, they'll be fixed in https://github.com/Kong/developer.konghq.com/pull/355

app/_gateway_entities/target.md
```diff
@@ -23,11 +23,16 @@ schema:
...
+    - gateway/admin-oss
+    - gateway/admin-ee
+    - konnect/control-planes-config
+
 ---
 
 ## What is a Target?
 
-A Target is an IP address/hostname with a port that identifies an instance of a backend service.
+A Target identifies an instance of a service application using an IP address or hostname with a port.
```
cloudjumpercat, line 35:
```suggestion
A Target identifies an instance of a service application using an IP address or hostname with a port. A service application is an application you own that serves requests and responses.
```
I think what was missing for me was defining what a service application is with an example of sorts. I don't know if my rewrite is accurate.

app/_gateway_entities/route.md
```diff
@@ -45,31 +51,36 @@ Routes fulfill two responsibilities in {{ site.base_gateway }}:
...
 2. Use plugins to transform the request/response proxied using this Route
 
-A Route must be attached to a [Service](/gateway/entities/service/), and _may_ have one or more [Plugin](/gateway/entities/plugin/) entities attached.
+A Route must be attached to a [Gateway Service](/gateway/entities/service/), and _may_ have one or more [Plugin](/gateway/entities/plugin/) entities attached.
 
-## Route and Service interaction
+## Route and Gateway Service interaction
 
-Routes, in conjunction with [Services](/gateway/entities/service/), let you expose your services to applications with {{site.base_gateway}}. Routes also allow the same service to be used by multiple applications and apply different policies based on the route used.
+Routes, in conjunction with [Gateway Services](/gateway/entities/service/), let you expose service applications to clients with {{site.base_gateway}}.
```
cloudjumpercat, line 58:
```suggestion
Routes, in conjunction with [Gateway Services](/gateway/entities/service/), let you expose service applications to clients with {{site.base_gateway}}. A service application is an application you own that serves requests and responses.
```

app/_gateway_entities/service.md
```diff
@@ -34,11 +34,12 @@ api_specs:
 
 ## What is a Gateway Service?
 
-A Gateway Service represents your actual backend API or microservice.
+A Gateway Service represents a service application in your system.
```
cloudjumpercat, line 37:
```suggestion
A Gateway Service represents a service application in your system. A service application is an application you own that serves requests and responses.
```

app/_gateway_entities/upstream.md
```diff
@@ -32,12 +32,18 @@ related_resources:
...
+    - gateway/admin-oss
+    - gateway/admin-ee
+    - konnect/control-planes-config
 ---
 
 ## What is an Upstream?
 
-An Upstream refers to the service applications sitting behind {{site.base_gateway}}, to which client requests are forwarded. 
-In {{site.base_gateway}}, an Upstream represents a virtual hostname and can be used to [health check](/gateway/health-checks-circuit-breakers/), [circuit break](/gateway/health-checks-circuit-breakers/), and [load balance](#load-balancing-algorithms) incoming requests over multiple [Gateway Services](/gateway/entities/service/). In addition, the Upstream entity has more advanced functionality algorithms like least-connections, consistent-hashing, and lowest-latency.
+An Upstream enables load balancing by providing a virtual hostname and collection of [Targets](/gateway/entities/target/), or service application instances, to which client requests are forwarded.
```
cloudjumpercat, line 44:
```suggestion
An Upstream enables load balancing by providing a virtual hostname and collection of [Targets](/gateway/entities/target/), or service application instances, to which client requests are forwarded. A service application is an application you own that serves requests and responses.
```

cloudjumpercat (changes requested):
The only feedback I had is to define what a "service application" is and a general example. I had a hard time visualizing what this would be exactly.

lena-larionova:
> The only feedback I had is to define what a "service application" is and a general example. I had a hard time visualizing what this would be exactly.

I disagree with this feedback, but would like @rspurgeon 's take on it. The point of switching to the term "service application" is that we _shouldn't_ define it, it's simply a standardized industry term. It replaces "backend service", "upstream service", etc.

rspurgeon:
> > The only feedback I had is to define what a "service application" is and a general example. I had a hard time visualizing what this would be exactly.
> 
> I disagree with this feedback, but would like @rspurgeon 's take on it. The point of switching to the term "service application" is that we _shouldn't_ define it, it's simply a standardized industry term. It replaces "backend service", "upstream service", etc.

A suggestion to consider... We could briefly expand on a "service application" in the Gateway Service entity box, there is a lot of room there. I think defining it once there is sufficient (IMO).

> Gateway Services represent the _service applications_ in your system. These applications are the business logic components in your system responsible for responding to requests. You will generally map one Gateway Service for each _service application_.

Maybe massage the language a bit and see if it fits well?

app/_gateway_entities/target.md
rspurgeon: For this page specifically, instead of defining the term, what if we updated the diagram to contain the term so it's clearer what we mean.   Right now the box on the far right in the diagram has the term "upstream targets", but those are the "service applications". It would be cool if the diagram could show that the service applications can be on different hosts, networks, etc... Right now they are just shown in a single box.

lena-larionova: Ah, I missed those, nice catch.

app/_gateway_entities/route.md
rspurgeon: IMHO this definition shouldn't be necessary

app/_gateway_entities/target.md
rspurgeon: Thanks! Same w/ the Upstreams page as well

lena-larionova:
> Gateway Services represent the service applications in your system. These applications are the business logic components in your system responsible for responding to requests. You will generally map one Gateway Service for each service application.
> Maybe massage the language a bit and see if it fits well?

I had to break it up a bit, it is too long. The last sentence fits better in the general description directly on the page rather than the tile (makes the tile too long and provides too much background for an at-a-glance description).

cloudjumpercat (approved):

PR #595, opened 2025-03-04
Add page title to TOC

mheap:
## Description

This is a proposal. How does the site look when the page title is in the TOC?

e.g. https://deploy-preview-595--kongdeveloper.netlify.app/gateway/networking/dns-config-reference/

### DIFF
--- a/app/_includes/layouts/main.html
+++ b/app/_includes/layouts/main.html
@@ -1,4 +1,4 @@
-<main class="flex flex-col pb-12 gap-12 {{ page.content_type | slugify }} {{include.class}}">
+<main id="main" class="flex flex-col pb-12 gap-12 {{ page.content_type | slugify }} {{include.class}}">
     <div class="flex flex-col gap-3">
         <div class="flex gap-3 items-center">
             {% if page.icon %}<img src="{{ page.icon }}" class="w-8 h-8"/>{% endif %}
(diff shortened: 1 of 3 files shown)


lena-larionova:
I like the idea of a link back to top (and having a default link when there's no other headings), but something about this feels off. What if it was an "On this page" or "back to top" link? Or a "jump to", something like that?

fabianrbz:
I like it, I would remove the `Overview` h2 too.

mheap:
Screenshots from decK.

It doesn't show up if there are no other headings
CleanShot 2025-03-05 at 10 02 04

An example where the most important part will not be represented in the TOC without this change

CleanShot 2025-03-05 at 10 02 41

lena-larionova:
> An alternative is to have a 1-2 line intro, then a h2 on each page. Otherwise the TOC doesn’t indicate a big chunk of data.

Not a huge fan of this approach, it forces us into potentially overexplaining things and forcing content into a rigid structure that isn't always helpful. Agree with Fabian that the "Overview" h2 needs to go (we're removing it from all the plugins, for instance).

After some more thinking about this, I'm fine with page title. Maybe it could use some formatting in the nav - make it bold? But otherwise, let's merge it, test it, and we'll adjust if it doesn't work in whichever situations.

lena-larionova (approved):
Approving with the caveat for future design adjustments.

PR #777, opened 2025-03-14
Feat: Response rate limiting plugin

lena-larionova:
## Description

Migrate https://docs.konghq.com/hub/kong-inc/response-ratelimiting/

Notes on includes:
* Even though I had to add a new section to the include for strategies for this plugin, it seemed worth it so that everything is in one place. Could split it out though.
* Couldn't really use the headers include here, this plugin is meant for interpreting custom headers so the usual headers info doesn't work.

## Checklist 

- [x] Every page is page one.


app/_kong_plugins/response-ratelimiting/index.md
```diff
@@ -37,9 +36,73 @@ search_aliases:
...
+After adding the plugin, you can increment the configured limits by adding the following response header:
+
+```
+Header-Name: Limit=Value [,Limit=Value]
+```
+
+With the default header name, `X-Kong-Limit`, the request looks like this:
+
+```bash
+curl -v -H 'X-Kong-Limit: limitname1=2, limitname2=4'
```
lmilan, line 65: An example URL could be useful here

app/_kong_plugins/response-ratelimiting/examples/response-rate-limit-with-local-policy.yaml
```diff
@@ -0,0 +1,19 @@
+description: |
+  Configure a custom response limit named `example_limit_name` with 10 responses allowed per minute.
+
+title: 'Response rate limit with local policy'
+
+weight: 900
+
+config:
+  limits:
+    example_limit_name:
```
lmilan, line 10:
```suggestion
    example-limit-name:
```
When I tested the underscores were replaced by dashes in the header name, so it might make sense to use that directly

app/_kong_plugins/response-ratelimiting/index.md
```diff
@@ -37,9 +36,77 @@ search_aliases:
...
+Header-Name: Limit=Value [,Limit=Value]
+```
+
+With the default header name, `X-Kong-Limit`, the request looks like this:
+
+```bash
+curl http://localhost:8000/example-route -v -H 'X-Kong-Limit: limitname1=2, limitname2=4'
+```
+
+The above example increments the limit `limitname1` by 2 units, and `limitname2` by 4 units.
```
lena-larionova, line 72:
```suggestion
```

PR #979, opened 2025-04-02
Feat: Route Transformer Advanced plugin

### LINKED ISSUE #834: Route Transformer Advanced
https://docs.konghq.com/hub/kong-inc/route-transformer-advanced/

Need to turn https://docs.konghq.com/hub/kong-inc/request-transformer-advanced/how-to/templates/ into an include, and then it can be used across all the route/request/response transformer plugins.

lmilan:
## Description

Fixes #834 

## Preview Links

- https://deploy-preview-979--kongdeveloper.netlify.app/plugins/route-transformer-advanced/

## Checklist 

- [x] Every page is page one.
- [x] Every page has a `description` entry in frontmatter


app/_includes/plugins/request-response-transformer/templates.md
```diff
@@ -0,0 +1,90 @@
...
+The content of the placeholder `$(...)` is evaluated as a Lua expression, so
+logical operators may be used. For example:
+
+```
+$(uri_captures["user-id"] or query_params["user"] or "unknown")
+```
+
+This will first look for the path parameter named `user-id` (in `uri_captures`). If not found, it will
+return the query parameter named `user`. If that also doesn't exist, it returns the default
+value '"unknown"'.
```
lena-larionova, line 51:
```suggestion
This example will look for parameters in the following order:
1. Looks for a path parameter named `user-id` in `uri_captures`.
2. If not found, it will return the query parameter named `user`.
3. If that also doesn't exist, it returns the default value `"unknown"`.
```

app/_includes/plugins/request-response-transformer/templates.md
```diff
@@ -0,0 +1,90 @@
...
+
+Constant parts can be specified as part of the template outside the dynamic
+placeholders. For example, creating a basic-auth header from a query parameter
+called `auth` that only contains the base64-encoded part:
```
lena-larionova, line 55:
```suggestion
Constant parts can be specified as part of the template outside the dynamic
placeholders. For example, you can create a basic-auth header from a query parameter
called `auth` that only contains the base64-encoded part:
```

app/_includes/plugins/request-response-transformer/templates.md
```diff
@@ -0,0 +1,90 @@
...
+Basic $(query_params["auth"])
+```
+
+Lambdas are also supported if wrapped as an expression like this:
+
+```
+$((function() ... implementation here ... end)())
+```
+
+A complete Lambda example for prefixing a header value with `Basic` if it's not
```
lena-larionova, line 67:
```suggestion
Here's a complete Lambda example for prefixing a header value with `Basic` if it's not
```

app/_includes/plugins/request-response-transformer/templates.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+{:.info}
+> **Note:** Especially in multi-line templates like the example above, make sure not
+to add any trailing white space or new lines. Because these would be outside the
+placeholders, they would be considered part of the template, and hence would be
+appended to the generated value.
+The environment is sandboxed, meaning that Lambdas will not have access to any
+library functions, except for the string methods (like `sub()` in the example
+above).
```
lena-larionova, line 90:
```suggestion
The environment is sandboxed, meaning that Lambdas won't have access to any
library functions, except for the string methods (like `sub()` in this example).

{:.info}
> **Note:** Make sure not to add any trailing whitespace or newlines, especially in multi-line templates. 
These would be outside the placeholders and would be considered part of the template, and hence would be appended to the generated value.
```
The second half looks like it doesn't relate to the first part at all, and instead describes the example - so i moved it.

lena-larionova (approved):

PR #1485, opened 2025-05-14
Fix links to kong.conf sections and update references

fabianrbz:
## Description

Fix links to kong.conf sections and update references
Skip processing a heading's id attribute based on
data-skip-process-heading-id=true.
Update the indices to use it too, since they were an exception too.

Added `-section` to the kong.conf section headings, `vaults` was both a section and a param causing the original issue because both ended up having the same id.
I also updated all the links

## Checklist 

- [x] Every page is page one.
- [x] Every page has a `description` entry in frontmatter


lena-larionova (approved):
lgtm, and thanks for updating the existing links to sections as well

PR #1515, opened 2025-05-16
Fix Konnect prereq for KIC how-tos

mheap:
The KIC prereqs still showed using a `license.json` even with Konnect. This PR fixes them

[/kubernetes-ingress-controller/mtls/](https://deploy-preview-1515--kongdeveloper.netlify.app/kubernetes-ingress-controller/mtls/)


cloudjumpercat (commented):
I retested the mTLS how to with Konnect and ran into the following:
- In the final validate step, I got this instead of the `200`. I don't know if this is expected or if I configured something incorrectly:
   ```
   Welcome, you are connected to node orbstack.
   Running on Pod echo-58df97db59-gn2rx.
   In namespace kong.
   With IP address 192.168.194.7.
   ```
- Since the cert generation for Konnect is now in the prereq, I think [this section](https://github.com/Kong/developer.konghq.com/blob/d68982005eb6410c4ccd60c13c2fe2c1e6c13554/app/_how-tos/kic-plugin-mtls.md?plain=1#L54) can be conditionally rendered for `if works on='on prem'`

mheap:
That's the expected message. The HTTP return code is `200`.

For cert generation, they're two different certs. Konnect needs a CP/DP cert, and everyone needs an mTLS cert. I can see how that's confusing now that there are two cert generation steps in a single page

PR #1621, opened 2025-05-28
Fix: use public docker image for KNEP

lena-larionova:
Issue reported on Slack.

Variable is currently not working because jekyll doesn't seem to be parsing `event-gateway` correctly in 
`{{ site.data.event-gateway_latest.version }}`. 

These options work:
* `{{ site.data.products.event-gateway.releases[0].version }}`: super long and not sustainable, as the latest version won't be first in the array when we add the next one.
* Renaming `event-gateway.yaml` to `knep.yaml`: then `{{ site.data.knep_latest.version }}` works with no problem - but that's fairly invasive. Is that the only way to get the variable to work?
cc @fabianrbz

### DIFF
--- a/app/_data/products/event-gateway.yml
+++ b/app/_data/products/event-gateway.yml
@@ -1,6 +1,7 @@
 name: Event Gateway
 
 releases:
-  - release: "0.0"
+  - release: "0.1"
+    version: "0.1.0-beta"
     name: "v0"
     latest: true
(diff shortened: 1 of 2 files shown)


fabianrbz:
@lena-larionova you can access it using hash notation (the dots are just a shortcut), i.e.:
```
{{ site.data['event-gateway_latest'].version }}
```

fabianrbz (approved):

PR #1715, opened 2025-06-04
fix: Add that self-hosted gitlab cluster must be accessible 

Guaris:


app/service-catalog/integrations/gitlab.md
```diff
@@ -34,6 +34,7 @@ For each linked project, the UI can show a **Project Summary** with simple data
 
 * You need the [Owner GitLab role](https://docs.gitlab.com/ee/user/permissions.html) to authorize the integration. This is required for event ingestion.
 * Only [GitLab.com subscriptions](https://docs.gitlab.com/ee/subscriptions/gitlab_com/) are supported at this time.
+* If you're using a self-hosted GitLab instance, it must be accessible from the public internet or is otherwise reachable by {{site.konnect_short_name}}.
```
jeffyongtaotang, line 37: https://docs.konghq.com/konnect/network/#kong-gateway-hostnames
see if above url is worth to point out as ref? Danny mentioned this when she is reaching out to customer.

PR #1720, opened 2025-06-04
Fix Operator validation

Guaris:
```
if [[ ! $? -neq {{ expected.return_code }} ]]; then
echo "Did not receive the expected return code"
fi
```
-> 

```
if [[ $? -ne {{ expected.return_code }} ]]; then
  echo "Did not receive the expected return code"
fi
```


lena-larionova (approved):
Tested, successfully validated with both `-ne 0` and `-ne 1` on https://deploy-preview-1720--kongdeveloper.netlify.app/operator/dataplanes/get-started/hybrid/install/#install-crds. LGTM

PR #2011, opened 2025-07-01
OIDC diagram and example improvements

### LINKED ISSUE #1949: Add OIDC diagrams to their respective plugin examples
Two requests:
* OIDC diagrams are currently on the overview page, but the user was looking for them on the plugin examples and didn't think to look on overview at all. Should add diagrams to their relevant examples.
* On the overview page, the order of the diagrams does not match the order of the plugin precedence list: https://developer.konghq.com/plugins/openid-connect/#authentication-flows-and-grants

lena-larionova:
## Description

Fixes #1949 

* Adding auth method diagrams to their respective examples - the diagrams aren't new, they're completely unchanged, just moved into their own files. No need to review them.
* Adding links from the OIDC overview page sections to their respective examples and how-tos
* Re-ordering the auth methods on the OIDC overview page to reflect their precedence order

Reviewers:
I'm testing two different methods of including the diagrams for the examples:
* Show the diagram completely
* Hide the diagram in a collapsed section

If we show them, people won't miss the diagrams. If we hide them, they don't have to scroll as much, and looking at the diagram is completely optional. This really depends on where they're coming from - if they were linked from the OIDC overview page, they already saw the diagram and this is a repeat. If they're coming from anywhere else (search, google, oidc landing page), then the diagram is potentially new and relevant. Which approach do you prefer?

## Checklist 

### DIFF
--- a/app/_includes/plugins/oidc/diagrams/auth-code.md
+++ b/app/_includes/plugins/oidc/diagrams/auth-code.md
@@ -1,0 +1,50 @@
+<!--vale off-->
+{% mermaid %}
+sequenceDiagram
+    autonumber
+    participant client as Client <br>(e.g. mobile app)
+    participant kong as API Gateway <br>(Kong)
+    participant idp as IdP <br>(e.g. Keycloak)
+    participant httpbin as Upstream <br>(upstream service,<br> e.g. httpbin)
+    activate client
+    activate kong
+    client->>kong: HTTP request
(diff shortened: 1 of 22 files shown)


cloudjumpercat (commented):
Left one suggestion about precedence and including a sentence about what precedence means/how it works. This would be nice to have. I'll review this again when you implement the "show diagram by default" feedback from Slack (I think it's just session auth that's hidden?)

PR #2077, opened 2025-07-07
fix: correct the wrong description for 3.9 breaking changes

ms2008:
## Description

Correct the wrong description for 3.9 breaking changes

Before: `To import the decK configuration files that are exported from earlier versions`
After: `To import the decK configuration files that are exported from 3.9 series to earlier versions`

## Checklist 

### DIFF
--- a/app/gateway/breaking-changes.md
+++ b/app/gateway/breaking-changes.md
@@ -121,7 +121,7 @@ The `node_id` parameter is planned to be removed in 4.x.
 
 This release adds support for the Hugging Face provider.
 
-To import the decK configuration files that are exported from 3.9 series to earlier versions, use the following script to transform it so that the configuration file can be compatible with the latest version:
+To import the decK configuration files that are exported from the 3.9.x series to earlier versions of {{site.base_gateway}}, use the following script to transform it so that the configuration file can be compatible with the latest version:
 
 ```
 yq -i '(


oowl (approved):

fffonion (approved):

app/gateway/breaking-changes.md
```diff
@@ -114,14 +114,14 @@ Breaking changes in the 3.10.0.0 release.
...
-Manually specifying a `node_id` via Kong configuration (`kong.conf`) is deprecated. 
+Manually specifying a `node_id` via Kong configuration (`kong.conf`) is deprecated.
 The `node_id` parameter is planned to be removed in 4.x.
 
 #### AI Rate Limiting advanced plugin
 
 This release adds support for the Hugging Face provider.
 
-To import the decK configuration files that are exported from earlier versions, use the following script to transform it so that the configuration file can be compatible with the latest version:
+To import the decK configuration files that are exported from 3.9 series to earlier versions, use the following script to transform it so that the configuration file can be compatible with the latest version:
```
lena-larionova, line 124:
```suggestion
To import the decK configuration files that are exported from the 3.9.x series to earlier versions of {{site.base_gateway}}, use the following script to transform it so that the configuration file can be compatible with the latest version:
```

lena-larionova:
> > The phrasing suggests that you want to import decK files from 3.9.x to an earlier version
> 
> @lena-larionova Yes, this is exactly what I meant to express. The original description was incorrect. It's not about importing the old version into the new version, but rather importing the new version into the old one. It's very strange, and I don't understand why wrote such a strange record back then.

Very weird. In that case, made a tiny phrasing adjustment and approved.

PR #2120, opened 2025-07-10
AI Gateway known issue in 3.11

lena-larionova:
## Description

Will be fixed in an upcoming patch.

## Preview Links
https://deploy-preview-2120--kongdeveloper.netlify.app/gateway/changelog/#3-11-0-0-known-issues
https://deploy-preview-2120--kongdeveloper.netlify.app/gateway/breaking-changes/#known-issues-in-3-11-0-0

### DIFF
--- a/tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
+++ b/tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
@@ -1,0 +1,8 @@
+message: |
+  If the AI Gateway has been enabled for more than a week, upgrades from 3.10 versions to 3.11.0.0 will fail due to a license migration issue. 
+  A fix will be provided in 3.11.0.1.
+
+  See [breaking changes in 3.11](/gateway/breaking-changes/#known-issues-in-3-11-0-0) for a temporary workaround.
+
+type: "known-issues"
+scope: "Core"
--- a/tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
+++ b/tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
@@ -1,8 +1,22 @@
 message: |
-  If the AI Gateway has been enabled for more than a week, upgrades from 3.10 versions to 3.11.0.0 will fail due to a license migration issue. 
-  A fix will be provided in 3.11.0.1.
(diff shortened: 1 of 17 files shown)


tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
```diff
@@ -0,0 +1,8 @@
+message: |
+  If the AI Gateway has been enabled for more than a week, upgrades from 3.10 versions to 3.11.0.0 will fail due to a license migration issue.
```
fffonion, line 2: I'm not sure if we want to be more specific or more general in this wording.
If we want to go more specific, I would say `If any AI plugins has been enabled for more than a week` , or rather list the names of AI plugins too. And we might want to exclude Konnect users and only mention on-prem deployment.
If we want to go more generic, I'm fine with the above.

lena-larionova: > would say If any AI plugins has been enabled for more than a week , or rather list the names of AI plugins too. And we might want to exclude Konnect users and only mention on-prem deployment.

I'm good with more specific, if it's accurate. Then it's more clear which subset of users are affected.

lena-larionova: Adjusted + added it to the individual plugin changelogs as well.

fabianrbz (approved):

oowl (approved):

tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
oowl: If you want to be more accurate, the Enterprise Edition plugin is more accurate; CE plugin does not affect.

fffonion (approved):

tools/changelog-generator/missing_entries/3.11.0.0/ai-gateway-license-migration.yml
fffonion: I would suggest to not mention CE/EE here, as there may become ambigous on whether it's a CE only plugin or a CE deployment.

lena-larionova: This is a bit of a grey area, we're avoiding making that distinction right now. Better not to mention at the moment.

PR #2287, opened 2025-07-22
feat(sdk): automated oas update - from docs repo

kong-docs:
Generated OAS files from docs.konghq.com


Guaris:
This is the opposite of what we want.

PR #2421, opened 2025-07-31
Feat(tools): Comparison table for tools

### LINKED ISSUE #1871: Add "when should I use X?" section to tool pages
Users really liked the "when should I use terraform?"  on the [terraform landing page](https://developer.konghq.com/terraform/)
It helped them to quickly decide if terraform is what they needed.

**Definition of Done:**
Add "when should I use X?" on all the tool pages.

### LINKED ISSUE #1878: Clarify which platforms decK, TF, KIC support on the tools overview page
**Definition of Done:**

Clarify which platforms decK, TF, KIC support on the tools overview page, a table might be enough.

lena-larionova:
## Description

Fixes #1871 
Fixes #1878 

These two issues were very similar + overlapping. I also couldn't _exactly_ tell what issue 1878 was referring to when it said "platforms", but the user testing feedback mentioned something about "knowing what I can use these tools with".

One more note: This page covers konnect apps and gateway but not our other products. Can/should we also add Mesh and KGO? Is KGO a tool? Should we position it as one? 
* Should probably add a column in the table that lists what products these tools all work with


cloudjumpercat (approved):
Just one small suggestion! Since you're on PTO, I'll get the suggestion applied and merge the PR.

Issue #2473, opened 2025-08-06
Kong gateway operator docs missing ingress controller version


phanijes:
https://developer.konghq.com/operator/dataplanes/get-started/kic/create-gateway/#gatewayconfiguration

Kong ingress controller version is missing in the above example from Kong docs and causing below error.

error: error parsing gateway-config.yml: error converting YAML to JSON: yaml: line 20: mapping values are not allowed in this context

PR #2593, opened 2025-08-15
Fix: docs missing ingress controller version 

Guaris:
Fixes https://github.com/Kong/developer.konghq.com/issues/2473


cloudjumpercat (approved):
Ran the whole series, looks like the namespace `kong` is already created in the prereq. I've left a suggestion to remove the part where you added a step to create it. The whole series ran fine otherwise.

PR #2599, opened 2025-08-15
Feat(mesh):Example swap out of subchart docs

justindavies:
## Description

Initial go at the changes needed across the board to remove the reliance on the Postgres subchart after the Bitnami changes.

## Preview Links
https://deploy-preview-2599--kongdeveloper.netlify.app/gateway/install/kubernetes/on-prem/

## Checklist 


justindavies:
Took quite a while to navigate what will actually happen after the switchover for the Bitnami charts. Thankfully, postgres is one of the continued to be maintained charts even in the Secure images repo.  The OCI Helm chart will automatically reference the Secure charts after switchover. 

I still feel we should drop our reliance on the subchart as goodness knows what will happen in the future.  I'm also unsure if we would be breaking Broadcom's TOS by having this as a sub helm chart.

mheap (dismissed):
This works as copy/paste for me now, thank you. Two nitpicks:

1. We create the CP `values.yaml` file, then install the DB, then install the CP. We can rearrange the page to install the DB first, so that the CP values.yaml file makes sense when you see the DB connection details
2. `values.yaml` is too generic for the DB values file. Please rename to `values-database.yaml` or similar (like we use `values-cp` and `values-dp`

justindavies:
@Guaris we are good to merge now

Guaris (approved):

PR #2631, opened 2025-08-19
ko: update installation guides to new Helm chart and update references

pmalek:
## Description

Part of https://github.com/Kong/kong-operator/issues/1988

## Checklist 


lena-larionova (commented):
Looks good on the surface but I'd like to test these commands - just run through it quickly to make sure everything installs. Going to hold off approving until the version is available.

lena-larionova (approved):
Tested with cert-manager + nightly builds, this LGTM with the caveat that the cert-manager requirement will change so we shouldn't need any extra instructions.

PR #2635, opened 2025-08-19
fix(gateway/entities): Fix broken links and missing links in Gateway entities

lena-larionova:
## Description
Fixing broken and missing links that appear in Gateway entity examples.

Ran into issue when updating L1 menu docs. We have a lot of broken and inconsistently missing links in the entity examples. They don't get caught by the link checker because the API page doesn't 404, it just loads blank.

Also added `konnect-api` as a supported tool for Upstreams, it was missing.

### DIFF
--- a/app/_gateway_entities/upstream.md
+++ b/app/_gateway_entities/upstream.md
@@ -12,6 +12,7 @@ tools:
     - kic
     - deck
     - terraform
+    - konnect-api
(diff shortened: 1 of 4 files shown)


cloudjumpercat (approved):
A few links were incorrect, I think, but I've left suggestions for them.

PR #2695, opened 2025-08-25
Fix(KIC): Redirect from old docs website

Guaris:


app/_redirects
```diff
@@ -99,6 +99,7 @@
 /mesh/policies/meshgatewayroute /mesh/policies/meshhttproute/
 /mesh/policies/proxy-template  /mesh/policies/
 
+/kubernetes-ingress-controller/:version/installation/* /kubernetes-ingress-controller/install/
```
cloudjumpercat, line 102: I've checked older versions as well as latest, I can't see any /installation page that ever existed

cloudjumpercat (approved):
Approving since this shouldn't cause any harm even if `/installation` seems to not exist

PR #2704, opened 2025-08-26
Feat: plugin toolkit process doc

lena-larionova:
## Description

Following discussion in team sync, wrote this doc while generating plugin schemas in https://github.com/Kong/docs-plugin-toolkit/pull/161.

## Preview Links

No netlify preview needed. See https://github.com/Kong/developer.konghq.com/blob/feat/plugin-toolkit-doc/docs/docs-plugin-toolkit.md instead.

### DIFF
--- a/docs/docs-plugin-toolkit.md
+++ b/docs/docs-plugin-toolkit.md
@@ -9,7 +9,11 @@ Download and update all schemas using GitHub actions:
 
 1. (Optional) If you're adding a new plugin, you need to set up the directory for it before running any GitHub actions:
     1. Clone https://github.com/Kong/docs-plugin-toolkit.
-    1. Add a folder for the new plugin at `/schemas/<plugin-name>`.
+    1. Add a folder for the new plugin at `/schemas/<plugin-name>` with a `.gitkeep` file, so that the empty directory gets pushed to GitHub:
+        ```sh  
+        mkdir schemas/plugin-name
+        touch schemas/plugin-name/.gitkeep
+        ```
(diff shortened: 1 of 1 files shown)


docs/docs-plugin-toolkit.md
```diff
@@ -0,0 +1,70 @@
...
+Our plugin schemas are generated via [Kong/docs-plugin-toolkit](https://github.com/Kong/docs-plugin-toolkit). 
+Whenever there is a new plugin or changes to plugins in a release, we use this toolkit to update schemas for a particular release.
+
+## Download and update all plugin schemas (GH actions)
+
+Download and update all schemas using GitHub actions:
+
+1. (Optional) If you're adding a new plugin, you need to set up the directory for it before running any GitHub actions:
+    1. Clone https://github.com/Kong/docs-plugin-toolkit.
+    1. Add a folder for the new plugin at `/schemas/<plugin-name>`.
```
tomek-labuk, line 12: Just wondering, shouldn't it be two steps, as we have to tell the toolkit that the plugin exists. So what I did with the LLM Judge plugin was:

	```sh
	mkdir schemas/plugin-name
	touch schemas/plugin-name/.gitkeep
	```

tomek-labuk (changes requested):

docs/docs-plugin-toolkit.md
lena-larionova: Ah, nice catch. Usually I just make an empty json file here, but that works too - either one 😄

PR #2746, opened 2025-09-02
Feat(Insomnia): Landing page install tabs

### LINKED ISSUE #2694: Feat: Insomnia Install instructions
## Jobs to be done (optional)
A user should have the same experiencing when installing on Windows through https://developer.konghq.com/insomnia/#install-insomnia  as the other operating systems. 

## Definition of done
Update the insomnia landing page so that the user experience is similar to the dec install page:
https://developer.konghq.com/deck/#install-deck

* Delete [install on windows](https://developer.konghq.com/how-to/download-insomnia-for-windows/)
* Create a navtab experience that links to the download of the package
* Add redirects for install-on-windows

## Information

The goal for an application like Insomnia is to get a user to install it and use it as fast as possible, the windows installation package being locked behind a how-to guide creates a barrier to the installation that doesn't exist for linux or OSX.

lmilan:
## Description

Fixes #2694 

## Preview Links

- https://deploy-preview-2746--kongdeveloper.netlify.app/insomnia/#install-insomnia

## Checklist 

### DIFF
--- a/app/_includes/tools/insomnia/install/linux.md
+++ b/app/_includes/tools/insomnia/install/linux.md
@@ -1,0 +1,2 @@
+1. Go to the [Insomnia downloads page](https://insomnia.rest/download) and download the installer for Ubuntu.
+1. Run the installer.
--- a/app/_includes/tools/insomnia/install/osx.md
+++ b/app/_includes/tools/insomnia/install/osx.md
@@ -1,0 +1,2 @@
+1. Go to the [Insomnia downloads page](https://insomnia.rest/download) and download the installer for MacOS.
+1. Run the installer.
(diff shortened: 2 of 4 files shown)


Guaris:
I think this is closer to done. In the original issue I thought linking directly to the package for the operating systems would be stronger than just the install page. Is there a reason you chose to link directly to the install page instead? 

As I'm looking at it, maybe the solution is to bring back the OS cards, but instead of having the windows card link to a how-to guide, it could link to the direct link to the windows download like the OSX and Ubuntu ones do on the live site. That way we can reduce from two clicks, to one.

Guaris (approved):

PR #2772, opened 2025-09-04
feat(targets): Failover targets for 3.12

cloudjumpercat:
## Description

Fixes #issue

## Checklist 

### DIFF
--- a/app/_how-tos/route-requests-to-backup-targets.md
+++ b/app/_how-tos/route-requests-to-backup-targets.md
@@ -1,0 +1,131 @@
+---
+title: Route requests to backup Targets during failures
+content_type: how_to
+
+description: Use the Route by Header plugin to route requests based on a header value.
+
+products:
+    - gateway
+
+works_on:
+    - on-prem
+
+min_version:
+  gateway: '3.12'
+
+entities: 
+  - service
+  - route
+  - upstream
+  - target
+
+tags:
+  - failover
+search_aliases:
+  - backup target
+
+tldr:
+    q: How do I route requests to different Targets in case my other Targets are unhealthy?
+    a: "Create a Service, a Route, and an Upstream with either the `latency`, `least-connections`, or `round-robin` load balancing strategy. Configure primary targets on the Upstream with `failover: false` and a failover Target with `failover: true`."
+
+tools:
+    - deck
+related_resources:
+  - text: Upstream entity
+    url: /gateway/entities/upstream/
+  - text: Target entity
+    url: /gateway/entities/target/
+
+cleanup:
+  inline:
+    - title: Destroy the {{site.base_gateway}} container
+      include_content: cleanup/products/gateway
+      icon_url: /assets/icons/gateway.svg
+---
+
+## Start three Target backends
+
+In this tutorial, we're going to have two primary Targets and one [failover Target](/gateway/entities/target/#managing-failover-targets). In this configuration, traffic will be routed to only the primary Targets if they are both healthy. If they are *both* unhealthy, traffic will be routed to the failover Target.
+
+We'll use three backends to represent our three Targets using Docker’s ultra-small `hashicorp/http-echo` image:
+
+```sh
+docker run -d --rm --name primary1 -p 9001:5678 hashicorp/http-echo -text "PRIMARY-1"
+docker run -d --rm --name primary2 -p 9002:5678 hashicorp/http-echo -text "PRIMARY-2"
+docker run -d --rm --name failover -p 9003:5678 hashicorp/http-echo -text "FAILOVER"
+```
+
+## Configure a Gateway Service and Route
+
+Configure a Gateway Service and Route to point to the Upstream you created in the prerequisites:
+{% entity_examples %}
+entities:
+  services:
+    - name: example-service
+      host: example-upstream 
+      protocol: http
+  routes:
+    - name: example-route
+      paths:
+      - "/anything"
+      service:
+        name: example-service
+{% endentity_examples %}
+
+## Configure the Upstream Targets
+
+Now, you can configure Upstream with the two primary Targets `failover: false` and one failover Target `failover: true`. 
+
+{% entity_examples %}
+entities:
+  upstreams:
+    - name: example-upstream
+      targets:
+        - target: host.docker.internal:9001
+          weight: 100
+          failover: false
+        - target: host.docker.internal:9002
+          weight: 100
+          failover: false
+        - target: host.docker.internal:9003
+          weight: 50
+          failover: true
+{% endentity_examples %}
+
+## Verify the primary Targets handle traffic
+
+Run the following to verify that only the primary Targets handle traffic because they are both healthy:
+
+```sh
+for i in {1..10}; do curl -sS http://localhost:8000/anything; echo; done
+```
+
+You'll get an output like the following:
+```sh
+PRIMARY-2
+
+PRIMARY-1
+
+PRIMARY-1
+
+PRIMARY-2
+
+PRIMARY-2
+
+PRIMARY-2
+```
+{:.no-copy-code}
+
+## Mark primary Targets as unhealthy
+
+To mark the primary Targets as unhealthy, you can run the following:
+```sh
+docker stop primary1 primary2
+```
+
+## Validate
+
+You can now validate that since the primary Targets are *both* unhealthy, only the failover Target routes traffic:
+```sh
+for i in {1..6}; do curl -s http://localhost:8000/anything; echo; done
+```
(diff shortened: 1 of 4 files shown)


app/_how-tos/route-requests-to-backup-targets.md
```diff
@@ -0,0 +1,131 @@
...
+  - target
+
+tags:
+  - failover
+search_aliases:
+  - backup target
+
+tldr:
+    q: How do I route requests to different Targets in case my other Targets are unhealthy?
+    a: "Create a Service, a Route, and an Upstream with either the `latency`, `least-connections`, or `round-robin` load balancing strategy. Configure primary targets on the Upstream with `failover: false` and a failover Target with `failover: true`."
```
lena-larionova, line 29:
```suggestion
    a: "Create a Service, a Route, and an Upstream with one of the `latency`, `least-connections`, or `round-robin` load balancing strategies. Configure primary Targets on the Upstream with `failover: false` and a failover Target with `failover: true`."
```
I think "either...or" is just meant for two items.

app/_how-tos/route-requests-to-backup-targets.md
```diff
@@ -0,0 +1,131 @@
...
+          failover: false
+        - target: host.docker.internal:9002
+          weight: 100
+          failover: false
+        - target: host.docker.internal:9003
+          weight: 50
+          failover: true
+{% endentity_examples %}
+
+## Verify the primary Targets handle traffic
```
lena-larionova, line 95:
```suggestion
## Verify that the primary Targets handle traffic
```

app/_how-tos/route-requests-to-backup-targets.md
```diff
@@ -0,0 +1,131 @@
...
+
+Run the following to verify that only the primary Targets handle traffic because they are both healthy:
+
+```sh
+for i in {1..10}; do curl -sS http://localhost:8000/anything; echo; done
+```
+
+You'll get an output like the following:
```
lena-larionova, line 103:
```suggestion
You'll get an output like the following, where you can see the Targets cycling between `PRIMARY-1` and `PRIMARY-2`:
```

app/_how-tos/route-requests-to-backup-targets.md
```diff
@@ -0,0 +1,131 @@
...
+```sh
+docker stop primary1 primary2
+```
+
+## Validate
+
+You can now validate that since the primary Targets are *both* unhealthy, only the failover Target routes traffic:
+```sh
+for i in {1..6}; do curl -s http://localhost:8000/anything; echo; done
+```
```
lena-larionova, line 131: I think it's better to combine these two, since marking it unhealthy is only done for the purpose of validation.
````suggestion
## Validate failover

To validate that the failover Target works, let's mark the primary Targets as unhealthy by shutting down the hosts. Run the following:
```sh
docker stop primary1 primary2
```

You can now validate that since the primary Targets are *both* unhealthy, only the failover Target routes traffic:
```sh
for i in {1..6}; do curl -s http://localhost:8000/anything; echo; done
```
This time, the response should show the `FAILOVER` Target instead of `PRIMARY`.
````

File: app/_includes/gateway/failover-targets.md [hunk 5]
```diff
@@ -0,0 +1,14 @@
+<!-- used on /gateway/entities/upstream/ and /gateway/entities/targets/ -->
+You can set `failover: true` to indicate that a Target should be used as a failover (backup) Target in case the other, regular targets (`failover: false`) associated with an Upstream are unhealthy. The failover Target is only used when *all** regular Targets are unhealthy. Only the [latency](/gateway/entities/upstream/#latency), [least-connections](/gateway/entities/upstream/#least-connections), and [round-robin](/gateway/entities/upstream/#round-robin) load balancing algorithms use failover Targets.
```
lena-larionova, line 2:
```suggestion
You can set `failover: true` to indicate that a Target should be used as a failover (backup) Target in case the other, regular targets (`failover: false`) associated with an Upstream are unhealthy. The failover Target is only used when *all** regular Targets are unhealthy. 
Failover Targets are supported by the [latency](/gateway/entities/upstream/#latency), [least-connections](/gateway/entities/upstream/#least-connections), and [round-robin](/gateway/entities/upstream/#round-robin) load balancing algorithms. 
```
Rephrased from "only these algorithms support this" to sound a bit less negative.

lena-larionova (commented):
Ran through it, had no trouble.

Do we know if decK is going to support this before release? If not, then we'll need to use curl (and add a separate command to create an Upstream before creating Targets). 

If decK is going to support this soon, then we should mark the version of decK in the prereqs.

cloudjumpercat:
> Do we know if decK is going to support this before release? If not, then we'll need to use curl (and add a separate command to create an Upstream before creating Targets).

> If decK is going to support this soon, then we should mark the version of decK in the prereqs.

@lena-larionova I've asked eng about this and will adjust the docs depending on what I hear back.

locao (commented):
This looks pretty good!

I have just a few comments:
- I didn't see any mention that more than one target can be used as failover. If you think it's worth adding this info, you could say their weights will be respected when choosing which one will proxy the request.
- I see the Target Schema is included from somewhere else. Is it from the code? In the rendered doc it is missing the failover field.
- I think it is worth mentioning nothing blocks the user from adding a failover target to consistent-hashing and sticky-sessions upstreams, but they will be ignored, not used in regular balancing or acting as failover.

cloudjumpercat:
@locao Thanks for the feedback! I've applied changes for your suggestions. For this:

> I see the Target Schema is included from somewhere else. Is it from the code? In the rendered doc it is missing the failover field.

We autogenerate the schema closer to release in our `release/gateway-3.12` branch.

lena-larionova:
> @locao Thanks for the feedback! I've applied changes for your suggestions. For this:
> 
> > I see the Target Schema is included from somewhere else. Is it from the code? In the rendered doc it is missing the failover field.
> 
> We autogenerate the schema closer to release in our `release/gateway-3.12` branch.

We should be able to do this now - I'll get it updated on the release branch.

app/_how-tos/route-requests-to-backup-targets.md
```diff
@@ -0,0 +1,130 @@
...
+    url: /gateway/entities/upstream/
+  - text: Target entity
+    url: /gateway/entities/target/
+
+cleanup:
+  inline:
+    - title: Destroy the {{site.base_gateway}} container
+      include_content: cleanup/products/gateway
+      icon_url: /assets/icons/gateway.svg
+---
```
lena-larionova, line 44:
```suggestion

faqs:
  - q: Can I set more than one failover Target?
    a: Yes, Upstream support multiple failover Targets.
---
```
Something like this, to make sure that people don't think this tutorial is the only way.

lena-larionova (approved):
LGTM, and pre-approved. You can merge it with decK instructions if our confidence on that is high.

app/_includes/gateway/failover-targets.md
```diff
@@ -0,0 +1,16 @@
+<!-- used on /gateway/entities/upstream/ and /gateway/entities/targets/ -->
+You can set `failover: true` to indicate that a Target should be used as a failover (backup) Target in case the other, regular targets (`failover: false`) associated with an Upstream are unhealthy. The failover Target is only used when *all* regular Targets are unhealthy. You can set one or multiple failover Targets for an Upstream. 
+
+Failover Targets are supported by the [latency](/gateway/entities/upstream/#latency), [least-connections](/gateway/entities/upstream/#least-connections), and [round-robin](/gateway/entities/upstream/#round-robin) load balancing algorithms. If you set failover Targets for Upstreams with the [consistent-hashing](/gateway/entities/upstream/#consistent-hashing) or [sticky session](/gateway/entities/upstream/#sticky-sessions) load balancing algorithms they *won't* be used for regular load balancing or as a failover.
```
lena-larionova, line 4:
```suggestion
Failover Targets are supported by the [latency](/gateway/entities/upstream/#latency), [least-connections](/gateway/entities/upstream/#least-connections), and [round-robin](/gateway/entities/upstream/#round-robin) load balancing algorithms. 

{:.info}
> **Note**: If you set failover Targets for Upstreams with the [consistent-hashing](/gateway/entities/upstream/#consistent-hashing) or [sticky session](/gateway/entities/upstream/#sticky-sessions) load balancing algorithms, they *won't* be used for regular load balancing or as a failover.
```
Just saw this; splitting it out as a note as the limitation gets kind of lost in the paragraph.

app/_includes/gateway/failover-targets.md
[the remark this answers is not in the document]
cloudjumpercat: Fixed this in my latest commit!

PR #2837, opened 2025-09-10
ko: introduce KO produce name in jekyll config

pmalek:
## Description

Rename old `gateway_operator_product_name` to `operator_product_name` and set `operator_product_name` to `Kong Operator`.

This will allow easy renaming when we're ready to make the switch.

## Checklist 


pmalek:
> @pmalek I'm a bit confused about the goal of this PR. I know we're renaming Kong Gateway Operator to Kong Operator; I'm not sure what the purpose of this interim step is.
> 
> * Why replace all instances of `{{site.operator_product_name}}` with `{{site.gateway_operator_product_name}}` instead of just using `{{site.operator_product_name}}` with the definition `Kong Operator`?
> * Do we need to keep the old name around for any reason? Are we going to maintain `Kong Gateway Operator` and `Kong Operator` in parallel?

The reasoning here is that some pages explicitly call out KGO and should remain this way (e.g. migration guide from KGO to KO) while some or rather most refer to KGO and should have that renamed. This step simply introduces the second variable but another step down the line will rename KGO to KO in most of the guides (using the variable).

Does that sound good to you or would you see that done differently?

PR #2952, opened 2025-09-19
Fix(gateway): 3.12 redirects and tags

lena-larionova:


lena-larionova:
Missing redirects check is a false flag, it doesn't parse examples correctly. That'll be fixed separately.

PR #2958, opened 2025-09-22
Treat event gateway policies as plugins/policies

### LINKED ISSUE #2943: make event policies plugins

fabianrbz:
## Description

Fixes #2943

Similar to mesh policies and plugins, we now have support for `overview`, `examples` and `configuration reference` pages.
I've created one example for the decrypt policy, we need to define in which formats we want to render the config.

## Checklist 


lena-larionova:
When you click on an example, the URL in the address bar switches to Mesh:

PR #3067, opened 2025-10-01
Fix(Dev Portal): Rename API Builder -> API Catalog

Guaris:
Renaming the API Builder spec to API Catalog to align more with The Service Catalog integration.

XR https://github.com/Kong/platform-api/pull/1841


api-specs/konnect/api-catalog/v3/openapi.yaml
cloudjumpercat, line 1: I think we'll also need a corresponding change for both these updated files to https://github.com/Kong/platform-api/blob/main/.github/raise-pr-on-change.json

cloudjumpercat (approved):
LGTM! I think this will just need to wait to be merged until the L2 APIs menu change.

I think we'll also need a corresponding change for both these updated files to https://github.com/Kong/platform-api/blob/main/.github/raise-pr-on-change.json

miko-bautista:
@Guaris Anything left that's blocking this PR from merging? @jharmn and I were on a customer call this morning where this came up.

PR #3075, opened 2025-10-02
feat(knep-policy-example): render event gateway policy examples

### LINKED ISSUE #3002: event gateway policies config examples
## Definition of done

Event Gateway policies config examples are rendered similar to plugin examples for the 3 different phases using konnect-api for now.

fabianrbz:
## Description

Fixes #3002 

Render event gateway policy examples in the same way we render plugin examples

## Checklist 


mheap:
Can we also group the plugins on the index page in to:

* Cluster Policies
* Consume Policies
* Produce Policies

fabianrbz:
@mheap some support more than one type of `phase`, we could look into adding some filters like we do for gateway plugins

mheap:
Filters would work. My initial thought was to duplicate under each heading, but filters are a better choice

PR #3303, opened 2025-10-31
chore(event-gateway) Update links to the event gateway api

lena-larionova:
## Description

Release prep task.
* Moved the api files to their respective new locations in `app/_api/konnect/event-gateway` and `api-specs/konnect/event-gateway/`. Still using the old KNEP schema, but this way all we need to do is replace those files once we have the new ones.
* Updated all the links for schemas to `konnect/event-gateway`
* Updated all the doc links to the API to `/api/konnect/event-gateway/`
* Deleted old include for applying config (uses the old API, this method doesn't exist anymore)
* Moved get started guide out of `/how-tos/knep` into `/how-tos/event-gateway` - this is where all the new guides will live. Didn't move the two remaining guides since they're getting phased out (but leaving the files for now because I'm not 100% sure if we can still reuse info/pieces in other guides)

### DIFF
--- a/app/event-gateway/entities/listener.md
+++ b/app/event-gateway/entities/listener.md
@@ -22,12 +22,8 @@ works_on:
 tags:
   - policy
 
-# schema:
-#     api: event-gateway/
-#     path: /schemas/
-
-# api_specs:
-#     - konnect/event-gateway
+api_specs:
+    - konnect/event-gateway
 
 products:
     - event-gateway


lena-larionova:
@fabianrbz for the redirect check, it's failing on:
```
- Renamed app/_how-tos/knep/get-started-with-event-gateway.md, url: /event-gateway/get-started/
```
The problem is that the doc actually has the permalink `/event-gateway/get-started/`, so moving it doesn't need a redirect. Can we adjust the check to account for this in the future? No rush - skipping it for now, don't need it to work for this PR.

Edit: Nvm setting the ci label on the PR doesn't pass the check 😭

fabianrbz (approved):

PR #3419, opened 2025-11-06
Update aws-private-link.md for 2 new regions

smritikjaggi:
## Description

Fixes #issue

## Checklist 


Guaris (approved):
This looks good to me.  Though I do wonder as we add more regions, maybe we should split EU, AP, and US into individual tables so users can see all the options in one place.

healthy-pod (approved):

Issue #2301, opened 2025-07-23
Possible error in Serverless Gateway API


johnfitzpatrick:
Following these steps to create a [Serverless Gateway](https://developer.konghq.com/serverless-gateways/reference/#how-do-i-provision-a-serverless-gateway)

Step 2 doesn't work for me - the Proxy URL does not get created.  

The URL stated is "`curl -X PUT "$KONNECT_CONTROL_PLANE_URL/v2/cloud-gateways/configurations`".

I expect will resolve to like "`https://us.api.konghq.com/v2/cloud-gateways/configurations`"

When I create the Serverless Gateway through the UI, my browser makes a request to "https://global.api.konghq.com/v3/cloud-gateways/configurations" -  two differences are "**global**" and "**v3**"

<img width="941" height="342" alt="Image" src="https://github.com/user-attachments/assets/794154bd-07cb-4f2b-bb1c-efd989835219" />

When I make this request it works

```
curl -v -X PUT "https://global.api.konghq.com/v3/cloud-gateways/configurations" \
    -H "Accept: application/json"\
    -H "Content-Type: application/json"\
    -H "Authorization: Bearer $PAT" \
    --data '{
      "control_plane_id":  "'"$CONTROL_PLANE_ID"'",
      "control_plane_geo": "us",
      "dataplane_groups": [
        {
          "region": "na"
        }
      ],
      "kind": "serverless.v0"
    }'

```

PR #3831, opened 2026-01-06
fix(serverless): Incorrect setup command + missing step

lena-larionova:
## Description

Fixes https://github.com/Kong/developer.konghq.com/issues/2301

Fixing the serverless gateway setup instructions; the command to create a data plane was using the wrong endpoint.
Tested to make sure.

Also fixed inconsistent capitalization across the doc (Serverless Gateways, control plane, data plane).

Q for reviewers: why wouldn't this be a how-to? I know it's only two steps, but that hasn't stopped us before.

## Preview Links

https://deploy-preview-3831--kongdeveloper.netlify.app/serverless-gateways/reference/


lena-larionova:
Based on some new info, not turning this into a how-to for now. Serverless is going to change in the near future; will tackle that then.

PR #3961, opened 2026-01-21
fix(konnect): Add explanation for deprecated 'runtime group' term

lena-larionova:
## Description

Issue came up in Kapa & in previous Slack threads.

I can also put this into custom answers in Kapa directly, if we don't want this on the site.


lena-larionova:
Closing in favor of kapa custom answer.

PR #4049, opened 2026-01-30
fix(oidc-plugin): Explain why/how to use multiple client IDs

lena-larionova:
## Description

Answering the following question, which came through our AI assistant:

> why does the openid connect plugin allow multiple client_id parameters?

Answer is from Aapo on Slack.

## Preview Links
https://deploy-preview-4049--kongdeveloper.netlify.app/plugins/openid-connect/#multiple-clients

### DIFF
--- a/app/_kong_plugins/openid-connect/index.md
+++ b/app/_kong_plugins/openid-connect/index.md
@@ -594,9 +594,9 @@ curl -X GET "http://localhost:8000?client_id=2"
 ```
 
 {{site.base_gateway}} will look for the client ID in the following locations, in order of precedence:
-1. If [`config.client_arg`](./reference/#schema--config-client-arg) is set, {{site.base_gateway}} looks for that value in the request header, URI argument, and body (in that order).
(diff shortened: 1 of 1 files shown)


app/_kong_plugins/openid-connect/index.md
```diff
@@ -562,6 +562,45 @@ To enable DPoP for OpenID Connect:
...
+Or by its index value (starting with 1):
+
+```sh
+curl -X GET "http://localhost:8000?client_id=2"
+```
+
+{{site.base_gateway}} will look for the client ID in the following locations, in order of precedence:
+1. If [`config.client_arg`](./reference/#schema--config-client-arg) is set, {{site.base_gateway}} looks for that value in the request header, URI argument, and body (in that order).
+2. If `config.client_arg` is not set, {{site.base_gateway}} looks for a `client_id` in the request header, URI argument, and body (in that order).
+2. If no client is found in either of those places, {{site.base_gateway}} uses the first client ID and client secret pair.
```
Guaris:
```suggestion
1. If no client is found in either of those places, {{site.base_gateway}} uses the first client ID and client secret pair.
```

PR #4330, opened 2026-02-25
feat(kongctl): Command reference

lena-larionova:
## Description

We were missing all the kongctl command pages. This PR adds them using the same method as we use for decK.
* Generate all the cmd help into includes
* Generate reference pages from the includes
* Add index pages for all the top-level command groupings
* Add all of this to the kongctl index and landing page

Things I'm not sure about:
* I had to adjust the script that extracts help text to also look at the the commands + the sub-subcommands. Copilot did that work; the output of the script looks correct, but someone should review the script itself.

## Preview Links
https://deploy-preview-4330--kongdeveloper.netlify.app/index/kongctl/#cli-references - everything in this section.

### DIFF
--- a/app/kongctl/api/delete.md
+++ b/app/kongctl/api/delete.md
@@ -20,7 +20,7 @@ related_resources:
     url: /kongctl/api/
 ---
 
-Delete resources with `DELETE` using [{{site.konnect_short_name}} APIs](/apis/).
+Delete resources with `DELETE` using [{{site.konnect_short_name}} APIs](/api/).
 
 
 ## Command usage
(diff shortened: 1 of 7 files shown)


fabianrbz:
> Re: missing redirects check - deleted a file that was unnecessary; it was a duplicate of the kongctl landing page and had no reason to exist. There is no URL change.
Yeah, we can skip it for now. The problem was that there was a conflict, two files had the same url (before you deleted it) and only one was generating the actual page, which ended up being the landing page. It's a rare edge case that the check doesn't consider.

fabianrbz (approved):

PR #4426, opened 2026-03-05
Fix: Adjusted kongctl workflow and scripts to properly prune removed help usage pages

rspurgeon:
## Description

Updates workflows and scripts for kongctl such that it will first remove all help / usage text doc files before regenerating them such that we properly purge removed commands from our generated help documentation.


lena-larionova (approved):
LGTM - ran it locally and checked output/looked through generated pages, everything built successfully.

PR #4520, opened 2026-03-16
feat(event-gw): improve routing doc

### LINKED ISSUE #4215: [EGW 1.1] Support multiple clusters with a single certificate
## Jobs to be done (optional)

Introduces support for multiple virtual clusters from a single `*.<snisuffix>` SAN certificate.

## Definition of done
Description from spec:

> Configures DNS names assigned to brokers in virtual clusters.
> - `per_cluster_suffix` is the default and allocates one level in the hierarchy for virtual clusters: `broker-{node_id}.{virtual_cluster}.{sni_suffix}`
> - `shared_suffix` puts all brokers from every virtual clusters into the same level: `broker-{node_id}-{virtual_cluster}.{sni_suffix}`. This makes it easier to manage certificates for this listener.

Add something about this to the Forward to Virtual Cluster policy doc.
https://developer.konghq.com/event-gateway/policies/forward-to-virtual-cluster/

## Information
https://github.com/kong-konnect/gw-admin/pull/807
https://github.com/Kong/platform-api/pull/2097
https://konghq.aha.io/epics/EGTWY-E-34

## Due date (optional)

EGW 1.1 GA

## Size
S

jakubdyszkiewicz:
## Description

Introduces new `broker_host_format`. I also found this doc a bit confusing because
* It does not show how this is really configured. There's a link to the listener page, but I can't find the configuration there
* It does not explain anything on bootstrap port/hostname, which is enabled by default and should be used by the client
* This might clutter the architecture page now. If so, maybe we need to move it somewhere, but routing should be better explained IMHO with a real example. I could not find this in other place.

Fixes #4215

## Checklist 

  It does not touch how-to docs.
  It already does.
  No
  It seems so
  Please review


app/event-gateway/architecture.md
```diff
@@ -97,34 +101,127 @@ rows:
...
+type: event_gateway_policy
+policy_type: forward-to-virtual-cluster
+name: forward-sni
+data:
+  sni_suffix: .acme
+  advertised_port: 9092
+  broker_host_format: shared_suffix
+  type: sni
+{% endentity_example %}
+{% endif_version %}
```
lena-larionova: We don't have versioned docs anymore (other than APIs and schemas, essentially). The docs are now written as if evergreen, and if there's a version difference, that gets a callout.

````suggestion
{% new in 1.1 %} Alternatively, you can set `broker_host_format` to `shared_suffix` in the listener policy, so that you can use one wildcard SAN for all virtual clusters. In this case, the mapping looks like this:

```
bootstrap-my-event-gateway.acme:9092 → kafka1:9092 (bootstrap hostname)
broker-1-my-event-gateway.acme:9092 → kafka1:9092
broker-2-my-event-gateway.acme:9092 → kafka2:9092
broker-3-my-event-gateway.acme:9092 → kafka3:9092
```

In all cases, the client must also be able to resolve the hostnames to the IP address of the gateway.

Here is an example configuration. First, configure a listener with the port that you're using:

{% entity_example %}
type: listener
data:
  name: listener
  addresses:
  - 0.0.0.0
  ports:
  - 9092
{% endentity_example %}

Then configure an SNI forwarding policy:
{% entity_example %}
type: event_gateway_policy
policy_type: forward-to-virtual-cluster
name: forward-sni
data:
  sni_suffix: .acme
  advertised_port: 9092
  broker_host_format: shared_suffix
  type: sni
{% endentity_example %}
{% endif_version %}
````

lena-larionova (changes requested):

lena-larionova:
For the examples cluttering the architecture page, we should move them here: https://developer.konghq.com/event-gateway/policies/forward-to-virtual-cluster/examples/port-mapping/ (path: `/app/_event_gateway_policies/forward-to-virtual-cluster/examples`)- there's a port mapping and an SNI example in the policy itself. We can leave the old examples and then add two new ones for the new setting, then add some links from this architecture page to these examples.

PR #4541, opened 2026-03-17
feat(event-gw): backend cluster auth guide

### LINKED ISSUE #4217: [EGW 1.1] mTLS between gateway and the backend cluster
## Jobs to be done (optional)
Support mTLS between the gateway and the backend cluster, in the case that the backend cluster requires mTLS from clients.

## Definition of done
* New fields in backend cluster config - should add example
* config example?

## Information
https://konghq.atlassian.net/browse/EVG-5
https://konghq.aha.io/epics/EGTWY-E-17

## Due date (optional)

EGW 1.1 GA

## Size
L

jakubdyszkiewicz:
## Description

Fixes #4217

## Checklist 

  Tested on dev, not on prod yet

### DIFF
--- a/.github/styles/base/Dictionary.txt
+++ b/.github/styles/base/Dictionary.txt
@@ -496,6 +496,7 @@ Kustomize
 Kustomize
 kustomize
 kustomize
+keystores
 kv
 Lakera
 langchain


app/_how-tos/event-gateway/configure-backend-cluster-auth.md
```diff
@@ -0,0 +1,464 @@
...
+    - event-gateway
+
+works_on:
+    - konnect
+
+tags:
+    - event-gateway
+    - kafka
+
+description: "Configure SASL/PLAIN and mTLS authentication so that {{site.event_gateway}} can connect to a secured Kafka cluster."
```
lena-larionova: Q: If you were looking for this info as someone trying to set up auth, would you be looking for a specific method (ie, SASL/PLAIN or mTLS specifically)? I think we should split this guide into two docs - we can use most of the same content but just have one guide that does SASL/PLAIN, and one that does mTLS. That also makes it easier for both linking purposes (eg customer wants to set up mTLS auth and doesn't care about the other options), and for AI (bot could get confused if they think both methods are necessary for the workflow).

PR #4661, opened 2026-03-27
swap dalle for latest openai image model

jbaross:
## Description

OpenAI has deprecated the dalle models so we should update our examples to use the current image models. The underlying APIs for image generation have not changed so this should be a drop in replacement and I have not tested the new model.

Kappa issue: https://app.kapa.ai/93b55f65-cffb-4d0b-9da7-c8915a838dce/conversations/84bb7674-f729-4b53-8a84-7ed73dcce72b


tomek-labuk (approved):

fabianrbz (changes requested):
we need to update the plugin examples descriptions too, they mention the model

fabianrbz (approved):

PR #4693, opened 2026-04-01
feat(expressions): add raw string literals usage

dndx:
It seems that the documentation for raw string literal support in expressions got lost during the transfer from the old page, added it back on so user is aware of this useful tool.

## Description

- Fixed wording: "repeated escape becomes" → "repeated escaping becomes"
- Replaced inline code snippets with fenced code blocks for the before/after predicate examples, ensuring they render correctly on separate lines

## Checklist 


lena-larionova (approved):
I moved the changes into an include file, as the table was moved there since it's reused between the routing page and the plugin conditional expressions page. LGTM

PR #4815, opened 2026-04-10
docs(Support KB): How-to-check-the-behavior-of-retries-attribute.md

kong-kiransharma:
first article to test and document the end to end process.


kong-kiransharma (commented):
modified the format to see if the checks succeed

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
+---
+# REQUIRED: The title that appears at the top of the page
+title: About retries attribute of service object
```
Guaris: This title needs work. Maybe you could name it the same as it is in the Support KB.

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
...
+
+# REQUIRED: Must be set to "support" for support articles
+content_type: support
+
+# OPTIONAL: A brief description of what this support article covers
+# This appears in search results and meta tags
+description: About retries attribute of service object.
```
Guaris: Because the description is optional, it's almost better to not add it then to just copy the title. I think using AI here would be very beneficial.

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
...
+  - konnect
+
+
+# OPTIONAL: Related resources that provide additional context
+# Each entry should have a text label and a url
+related_resources:
+  - text: Link text for related documentation
+    url: /path/to/related/page/
+  - text: Another related resource
+    url: /path/to/another/page/
```
Guaris: This links or to be removed.

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
...
+
+# OPTIONAL: TL;DR section that appears at the top of the article
+# Provides a quick question and answer summary
+tldr:
+  q: retries attribute of service object
+  a: |
+    The articles describe the end-to-end steps of working with service object
+
```
Guaris: Think of this as a Question then an answer. So 

Q: How do I check the behavior of the retries attribute
A: When an upstream is unreachable, Kong retries the connection up to the configured number of times (e.g., retries=5). You can observe this in debug logs.

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
...
+---
+
+Below is an example of installing Kong on Docker.
+
+Please install Kong on Docker by following the official guide: Kong Docker Installation .
```
Guaris: This was copy and pasted from the original. Do you think we need it?

docs/support/About-retries-attribute-of-service-object.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+## 2. Does the service object only retry when an upstream object is set?
+
+No, the service object can retry requests even if an upstream object has not been set.
+
+## 3. Does the service object retry on HTTP 4xx/5xx responses?
+
+No, retries happen only for TCP connection errors. However, Kong can perform HTTP and TCP health checks using the upstream object. For more details, see the Health Checks and Circuit Breakers documentation .
+
```
Guaris: You can add these to the front matter using the following syntax: 
```
faqs:
  - q: What happens when an admin doesn't have a role assigned?
    a: If an admin is in a Workspace without a role, they can’t see or interact with anything. Admins can manage entities inside Workspaces, including users and their roles.
  - q: What is a super admin?
    a: A super admin is a role that has the ability to assign and modify RBAC roles and permissions. A generic admin without this role can't manage RBAC.
  - q: How can I invite and manage admins in Kong Manager?
    a: If you want to manage admins from the Kong Manager UI, go to Teams > Admins. From here, you can invite new admins, manage existing admins, and find invitation links for invited admins.
  - q: Can admins manage multiple Workspaces?
    a: No, each admin is specific to one Workspace.
```

Guaris (changes requested):

kong-kiransharma (commented):
modified with the comments.

kong-kiransharma:
@Guaris I made the suggested changes but I don't think my rendering of faq section is right. I played around with spacing & new line but it wasn't getting any nicer. Perhaps you could look into this for me once.

kong-kiransharma (commented):
.

kong-kiransharma (commented):
.

Guaris (approved):
test

PR #4869, opened 2026-04-15
Fix(Insomnia): update with troubleshooting steps for timeouts only in Insomnia

### LINKED ISSUE #4852: Insomnia: document request timeouts troubleshooting
## Jobs to be done (optional)

Add troubleshooting info on request timeouts when they only happen in Insomnia (not HTTPS, for example).

The root cause might be the `Content-Lenght` parameter in Insomnia, forcing the server to validate the content length against the body, and failing in case it mutates.

Steps to troubleshoot it:

1. Remove the `Content-Lenght` parameter from  the Mock Header tab.
2. Click **Test** to see if the issue is solved.

## Definition of done

The troubleshooting info is in the appropriate section, easy to find and Kapa uses it to help users troubleshoot the issue.
 
## Size

S

juliamrch:
## Description

This is a long-lasting issue (+2 years) that I didn't see documented anywhere, and for which Kapa wasn't able to provide a satisfactory answer to one of our users.

I documented the root cause and the troubleshooting steps. Not sure about the choice of the page, lmk if you think another page would be more appropriate.
 
Closes #4852

## Preview Links

https://deploy-preview-4869--kongdeveloper.netlify.app/insomnia/requests/#faqs

## Checklist 

### DIFF
--- a/app/insomnia/requests.md
+++ b/app/insomnia/requests.md
@@ -211,5 +211,5 @@ If you encounter request timeouts errors (400) only in Insomnia, the `Content-Le
 
 To resolve the error, try the following:
 
-1. From your request's **Mock Header** tab, remove or deactivate the `Content-Lenght` parameter.
+1. From your request's **Mock Header** tab, remove or deactivate the `Content-Length` parameter.
 2. Click **Test** to see if the issue is resolved.
--- a/app/insomnia/requests.md
+++ b/app/insomnia/requests.md
@@ -204,3 +204,12 @@ See the following pages for the CLI flag and configuration options:
 
 - [Inso CLI overview](/inso-cli/)
 - [Inso CLI configuration](/inso-cli/configuration/)
+
+### Troubleshooting
+
+If you encounter request timeouts errors (400) only in Insomnia, the `Content-Lenght` parameter might be forcing the server to validate the content length against the body, and failing in case it mutates.
+
+To resolve the error, try the following:
+
+1. From your request's **Mock Header** tab, remove or deactivate the `Content-Lenght` parameter.
+2. Click **Test** to see if the issue is resolved.
--- a/app/insomnia/requests.md
+++ b/app/insomnia/requests.md
@@ -63,6 +63,14 @@ faqs:
       {% raw %}https://api.example.com/users/{{ user_id }}{% endraw %}
       ```  
       For detailed usage examples, go to [**Pre-request scripts**](/how-to/write-pre-request-scripts/). 
+  - q: Why does my request return an error in Insomnia but succeed with other tools?
+    a: |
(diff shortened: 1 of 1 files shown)


app/insomnia/requests.md
```diff
@@ -204,3 +204,12 @@ See the following pages for the CLI flag and configuration options:
 
 - [Inso CLI overview](/inso-cli/)
 - [Inso CLI configuration](/inso-cli/configuration/)
+
+### Troubleshooting
+
+If you encounter request timeouts errors (400) only in Insomnia, the `Content-Lenght` parameter might be forcing the server to validate the content length against the body, and failing in case it mutates.
```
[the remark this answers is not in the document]
Guaris: @juliamrch Is this copilot suggestion worth looking into?

app/insomnia/requests.md
```diff
@@ -204,3 +204,12 @@ See the following pages for the CLI flag and configuration options:
...
+
+To resolve the error, try the following:
+
+1. From your request's **Mock Header** tab, remove or deactivate the `Content-Lenght` parameter.
```
Guaris:
```suggestion
1. From your request's **Mock Header** tab, remove or deactivate the `Content-Length` parameter.
```

app/insomnia/requests.md
Guaris:
```suggestion
If you encounter request timeouts errors (400) only in Insomnia, the `Content-Length` parameter might be forcing the server to validate the content length against the body, and failing in case it mutates.
```

Guaris:
```suggestion
If you get a `400` request timeouts error in Insomnia, but the same request works with cURL, it may be due to the `Content-Length` header. Insomnia might be sending a fixed `Content-Length`, causing the server to validate it against the body, and failing if they don't match.
```

Guaris (commented):
I would add this to the FAQ section of the same doc. I think a troubleshooting section would be a good addition to this doc if we have more than one type of issue, so like: 

```
## Troubleshooting
This is a list of common troubleshooting issues and their fixes

### Timeouts appear only in Insomnia

### Other issue 

```


What do you think?

app/insomnia/requests.md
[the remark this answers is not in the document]
juliamrch: Yes, it's more accurate.

juliamrch:
> I would add this to the FAQ section of the same doc. I think a troubleshooting section would be a good addition to this doc if we have more than one type of issue, so like:
> 
> ```
> ## Troubleshooting
> This is a list of common troubleshooting issues and their fixes
> 
> ### Timeouts appear only in Insomnia
> 
> ### Other issue 
> ```
> 
> What do you think?

I considered the option, but I wasn't sure the FAQ would be the right place for troubleshooting this specific issue.

Guaris (approved):

PR #4958, opened 2026-04-21
Feat(Platform): Algolia instructions internal docs

Guaris:


cloudjumpercat (changes requested):
Let's pull the Algolia retrigger docs from the release checklist and just have a link to this separate doc instead (since Algolia recrawls aren't just for releases).

docs/algolia.md
```diff
@@ -0,0 +1,11 @@
+# Algolia
+
+## Recrawl the site
+
+1. Log into Algolia using the credentials in 1pass
```
lena-larionova:
```suggestion
1. Log into Algolia using team-docs (credentials in 1pass).
```
There are two accounts in 1pass, this is the one we need to use.

cloudjumpercat (approved):

PR #4974, opened 2026-04-22
feat(gateway-manager): Konnect Workspaces

### LINKED ISSUE #4242: Gateway Manager: Konnect Workspaces
## Definition of done
- how to guide that explains how to configure a Konnect workspace (use draft here https://docs.google.com/document/d/1-oAYHB78YrEQNry3Lpanff85SlptilHxiAKR2604txU/edit?tab=t.0)
- update FAQ and link to how to: https://developer.konghq.com/gateway/entities/workspace/#can-i-use-workspaces-in-konnect
- Fix the frontmatter so this doesn't say incompatible with Konnect https://developer.konghq.com/gateway/entities/workspace/
- Verify if everything in [here](https://developer.konghq.com/gateway/entities/workspace/) applies to Konnect, adjust any content or create a new section as necessary

## Information
DRP: Ed B.
Slack: https://kongstrong.slack.com/archives/C09PFG2R1R8
PRD: https://docs.google.com/document/d/1iVO7uKxmCnMn1Q7UczE4sJJuy9Bihu3dqxiR_fWxX1o/edit?tab=t.4htgtiafh
Eng brief: https://docs.google.com/document/d/1S1fCOFdmNOLwaJCRvmypx8aoPXuzUjB8Fhy0_Ads-v8/edit?tab=t.0#heading=h.6bgidzgnv6ul

## Due date (optional)

GA mid March 

## Size
L

cloudjumpercat:
## Description

Fixes #4242 

**TODO:**

- [ ] For application <> consumer mapping: "With the introduction of workspaces, Dev Portal app registration continues to work at the control plane level, and Phase 1 supports only default-workspace consumers. Multi-workspace support for Dev Portal consumer linking is deferred to a later phase."
- [ ] On principals reference: "we should describe that you can map at most one Consumer per Workspace per Control Plane to a Principal. You can map multiple Consumers if they are in different Control Planes, or in the same Control Plane but a different Workspace."

## Preview Links
- https://deploy-preview-4974--kongdeveloper.netlify.app/gateway/entities/workspace/
- https://deploy-preview-4974--kongdeveloper.netlify.app/how-to/manage-konnect-workspaces/
- https://deploy-preview-4974--kongdeveloper.netlify.app/gateway/self-managed-migration/#migrating-from-self-managed-workspaces-to-konnect

## Checklist 


app/_gateway_entities/workspace.md
```diff
@@ -49,11 +49,15 @@ api_specs:
 
 works_on:
   - on-prem
+  - konnect
 
 tags:
   - rbac
 ---
 
+{:.important}
```
cloudjumpercat:
```suggestion
{:.warning}
```

app/_gateway_entities/workspace.md
```diff
@@ -49,11 +49,15 @@ api_specs:
...
 works_on:
   - on-prem
+  - konnect
 
 tags:
   - rbac
 ---
 
+{:.warning}
+> **{{site.konnect_short_name}} Workspaces are in public beta**
```
edbyford: Actually, Phase 1 is gated behind feature flags - customers should contact their Kong reps to get us to enable it for them, something like:

"Konnect Workspaces are in private beta. To use this feature, contact your Kong representative or Kong Support."

app/_gateway_entities/workspace.md
```diff
@@ -119,9 +123,10 @@ config:
 
 ## Roles, groups, and permissions
 
-Because Workspaces allow users to control {{site.base_gateway}} entities in isolation, users must have the correct permissions to configure a particular Workspace. Users will require either a Super Admin or Admin role to configure Workspaces. 
+Because Workspaces allow users to control {{site.base_gateway}} entities in isolation, users must have the correct permissions to configure a particular Workspace. {{site.base_gateway}} users will require either a Super Admin or Admin role to configure Workspaces. 
+{{site.konnect_short_name}} users will need a [Workspace Admin role](/konnect-platform/teams-and-roles/#konnect-workspaces) to configure Workspaces.
```
edbyford: This role does not exist yet. Workspace-level RBAC depends on KAuth nested permissions work, which is planned post-Phase 1 and blocks Phase 2.

Suggest adding wording to the effect of:

"To configure Workspaces in Konnect during the beta, users need the Control Plane Admin role on the parent control plane. Workspace-scoped roles will be introduced in a future phase."

app/_gateway_entities/workspace.md
edbyford: The conflict detection algorithm described in lines 102–109, and the route_validation_strategy defaults pulled in by the config table, reflect on-prem Kong Gateway behaviour.
 
Konnect Workspaces Phase 1 only supports exact-match collision detection (path + method + host). The smart strategy is not available. 

We should probably add a Konnect-specific note here, or carve out a separate algorithm description for Konnect, so users don't expect smart-strategy behaviour. 

@aboudreault  will be able to provide more details if needed

edbyford (changes requested):
Three corrections flagged inline: beta status, Konnect Workspace Admin role doesn't exist yet, route validation default is exact-match not smart for Konnect. Happy to pair on wording.

app/_gateway_entities/workspace.md
aboudreault: Agree with Ed. For konnect it's with these fields: path + method + host + sni

edbyford:
The beta status contradicts itself between two places on the page:

- The warning callout near the top: "Konnect Workspaces are in private beta… contact your Kong representative or Kong Support."
- The FAQ answer to "Can I use Workspaces in Konnect?": "Yes. Konnect Workspaces are in public beta."

It's a private beta, enabled per organisation by Kong (not self-service), so the callout is the correct one. Could you update the FAQ answer to match? Something like:

> Yes, in private beta. Konnect Workspaces are enabled per organisation, so contact your Kong representative or Kong Support to request access.

One other thing while you're here (happy to spin this into a separate issue if you'd rather): the page covers the entity reference, the API/decK example snippets, and the UI create step, but there's no how-to or quickstart for the Konnect workspace lifecycle end to end (create a control plane, create a workspace, manage entities inside it, use decK with `--workspace` / `--all-workspaces`). Anyone setting up via the API or decK has no walkthrough to follow. Worth a short Konnect Workspaces how-to. I've recently verified the live API paths and decK behaviour against the beta, so happy to supply tested steps.

app/_how-tos/konnect-platform/manage-konnect-workspaces.md
```diff
@@ -0,0 +1,158 @@
...
+
+# Apply a config file into a Workspace
+deck gateway sync team-payments.yaml --konnect-control-plane-name my-cp --workspace team-payments
+```
+
+A dumped file carries a `_workspace:` header, so `deck gateway sync` will target the right Workspace even without the `--workspace` flag.
+
+## Route collisions
+
+When you create or modify a Route, {{site.konnect_short_name}} checks it against Routes in every Workspace on the control plane, including the `default` one. If it overlaps with an existing Route on path, method, host, and SNI, the request is rejected with a `400` naming the collision. This catches the cross-team conflicts that may be introduced when several teams share a data plane.
```
edbyford, line 182: Small fix: this should be `409`, not `400`. Konnect was returning `400` for route collisions until 11 May, when the team fixed the internal Postgres error mapping. Verified end-to-end against `.tech` on 19 May - the response is now `HTTP 409 {"code":6,"message":"route collision for paths: /orders"}`, matching the on-prem behaviour and the wording already on the `gateway/entities/workspace/` page.

app/_how-tos/konnect-platform/manage-konnect-workspaces.md
```diff
@@ -0,0 +1,158 @@
...
+<!--vale on-->
+
+{:.info}
+> Entities in the `default` Workspace use the existing un-prefixed path, for example `.../core-entities/services`.
+
+## Use decK
+
+To use decK with {{site.konnect_short_name}} Workspaces, you need 1.43 or later, but 1.60 or later is recommended.
+
+```shell
```
edbyford: Follow-up from the Slack thread: when Diana ran the commands above (and when I tried them) without the right env vars set, decK fell back to on-prem admin-API mode and threw:

```
Error: reading Kong version: making HTTP request: Get "http://localhost:8001/team-payments/kong": dial tcp [::1]:8001: connect: connection refused
```

The commands as written only have `--konnect-control-plane-name`, so they rely on env vars for the token and the address. decK picks them up only with the `DECK_` prefix - `DECK_KONNECT_TOKEN`, `DECK_KONNECT_ADDR`. A bare `KONNECT_TOKEN` doesn't work.

Worth adding a short setup block just before the code block, something like:

```shell
export DECK_KONNECT_TOKEN="kpat_..."                 # your Konnect PAT
export DECK_KONNECT_ADDR="https://us.api.konghq.com" # or .tech if your org lives there
```

Then the commands below work as written.

edbyford: also it will need deck "1.57 or later, but 1.60 or later is recommended"

edbyford:
Could you add a short **custom plugins** section? Going into 3.15 custom plugins are being de-workspaced - schemas live at the control plane level only, instances live per workspace, same as bundled plugins (and same model as on-prem). Worth a heads-up for migrating customers.

Suggested new H2 between "Manage entities in a Workspace" and "Use decK":

## Custom plugins

Install custom plugin schemas at the control plane level (in `default`). Once installed, you can create plugin instances of that custom plugin in any workspace on the control plane, the same way you would with a bundled plugin.

```sh
# schema lives in default
curl -X POST .../core-entities/custom-plugins ...

# instance can live in any workspace
curl -X POST .../core-entities/$WORKSPACE/plugins \
  -d '{ "name": "my-custom-plugin", "config": { ... } }'
```

Installing schemas in a non-default workspace is not supported.

cloudjumpercat:
> Could you add a short **custom plugins** section? Going into 3.15 custom plugins are being de-workspaced - schemas live at the control plane level only, instances live per workspace, same as bundled plugins (and same model as on-prem). Worth a heads-up for migrating customers.
> 
> Suggested new H2 between "Manage entities in a Workspace" and "Use decK":
> 
> ## Custom plugins
> Install custom plugin schemas at the control plane level (in `default`). Once installed, you can create plugin instances of that custom plugin in any workspace on the control plane, the same way you would with a bundled plugin.
> 
> ```shell
> # schema lives in default
> curl -X POST .../core-entities/custom-plugins ...
> 
> # instance can live in any workspace
> curl -X POST .../core-entities/$WORKSPACE/plugins \
>   -d '{ "name": "my-custom-plugin", "config": { ... } }'
> ```
> 
> Installing schemas in a non-default workspace is not supported.

@edbyford Thanks! I've added this in my latest commit.

edbyford:
Sorry for the late tweak — Fero flagged that the "same model as on-prem" parallel doesn't quite land, since custom plugins were disabled by default on-prem and there were no Konnect workspaces before, so neither side of the comparison gives readers a useful anchor.

Could we adjust the intro paragraph to drop both parallels? Suggested wording:

> In 3.15, custom plugin schemas live at the control plane level and instances live per workspace, the same as bundled plugins. If you're already running custom plugins on a Konnect control plane, the schemas stay where they are. Existing plugin instances remain in the `default` workspace — to scope a plugin to a different workspace, create a new instance inside that workspace.

The code block underneath can stay as-is. Happy to raise a follow-up PR if easier.

app/_how-tos/konnect-platform/manage-konnect-workspaces.md
```diff
@@ -0,0 +1,183 @@
...
+- **Configurable Route collision strategies.** The behavior described in [Route collisions](#route-collisions) is fixed in this phase. 
+- **Terraform provider support for Workspaces.** 
+
+### {{site.konnect_short_name}} Workspace limitations
+
+The following are {{site.konnect_short_name}} Workspace limitations:
+
+- A Workspace-enabled control plane cannot be part of a Control Plane Group.
+- Workspaces are not supported on serverless control planes.
+- [Route collision](#route-collisions) detection uses exact match, not the "smart" strategy available on-prem.
```
cloudjumpercat, line 57:
```suggestion
- [Route collision](#route-collisions) detection uses exact match, not the "smart" strategy available on-prem.
- You can have a maximum of 3,000 Workspaces in {{site.konnect_short_name}}
```

PR #5200, opened 2026-05-12
Feat(Platform): add token count to llm-mode and docs

Guaris:
Exposes an estimated token count beside the copy mode and in the text version of the docs. 

I got the idea from this [article](https://addyosmani.com/blog/agentic-engine-optimization/), where it says: 

```
This one is simpler than it sounds but surprisingly high-leverage: surface token counts on your documentation pages. Ideally in both the llms.txt index and on the pages themselves (as metadata or a page header).

This gives agents the information they need to make smart decisions:

“This page is 8K tokens - I can include it fully in context”
“This page is 150K tokens - I should fetch only the relevant section”
“This page exceeds my context window - I’ll use the summary from llms.txt instead
```

image
image


Guaris:
Killing because its subjective

PR #5206, opened 2026-05-12
fix(deck): Explain how DECK_ env variables work for Konnect

lena-larionova:
## Description

When asked "Does deck automatically check for KONNECT_TOKEN or DECK_KONNECT_TOKEN?", Kapa provides an uncertain half answer - the answer is correct, but Kapa has to infer it from our how-to guides, so it isn't completely sure. Adding info about connecting to Konnect with decK env variables explicitly into the docs so that it's findable + indexable.

### DIFF
--- a/app/deck/gateway/konnect-configuration.md
+++ b/app/deck/gateway/konnect-configuration.md
@@ -21,16 +21,19 @@ related_resources:
     url: /konnect/
   - text: deck gateway commands
     url: /deck/gateway/
+  - text: decK configuration
+    url: /deck/configuration/
 ---
 
 {:.warning}
 > **Note**: {{site.konnect_short_name}} requires decK v1.40.0 or above. Versions below this will see inconsistent `deck gateway diff` results.
 
-You can manage {{site.base_gateway}} core entity configuration in your {{site.konnect_short_name}} organization using decK.
+To use decK with {{site.konnect_short_name}}, you need:
+- A {{site.konnect_short_name}} authentication token
+- The name of a control plane to target
+- (Optional, if your organization isn't in the `us` region) The {{site.konnect_short_name}} API address
 
-decK can only target one control plane at a time.
-
-Managing multiple control planes requires a separate state file per control plane.
+decK can only target one control plane at a time. Managing multiple control planes requires a separate state file per control plane.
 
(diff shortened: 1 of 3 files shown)


app/deck/gateway/konnect-configuration.md
```diff
@@ -77,28 +79,66 @@ To target a specific geo, set `konnect-addr` to one of the following:
...
   ```
 
-- Set a Control Plane using the `--konnect-control-plane-name` flag:
+- Set a control plane using the `--konnect-control-plane-name` flag:
 
   ```sh
   deck gateway sync konnect.yaml --konnect-control-plane-name staging
   ```
 
+### Setting default connection values
```
mheap, line 105: https://developer.konghq.com/deck/configuration/

Is this duplicate content? Should the config page be fleshed out more, or merged into this page?

lena-larionova: I'd say this is Konnect-specific - but we could merge the pages.

app/deck/gateway/konnect-configuration.md
```diff
@@ -77,28 +79,66 @@ To target a specific geo, set `konnect-addr` to one of the following:
...
+1. Use decK without passing any of the connection values:
+
+   ```bash
+   deck gateway ping
+   ```
+
+   The response will show which {{site.konnect_short_name}} organization decK connected to:
+
+   ```sh
+   Successfully Konnected to the My Example Org organization!
```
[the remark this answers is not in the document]
lena-larionova: > so users don't think decK will always print their org name.

decK _will_  always print their org name on a `deck gateway ping`.

app/deck/gateway/konnect-configuration.md
mheap: I don't feel strongly either way. Was just flagging that the non-concrete version of this existed

lena-larionova: I'll think of how to combine/resolve; either way, we need to flesh this out somewhere, as it's not clear for a Konnect user that they don't have to pass all the flags every time.

mheap: This capability isn't Konnect specific. It works for other flags too

cloudjumpercat (approved):
LGTM!

app/deck/gateway/konnect-configuration.md
lena-larionova: Shortened the original change to a couple of lines and cross-linked to the other doc instead.

PR #5216, opened 2026-05-13
fix: improving the docker-compose page suggesting to create the kong

thesp0nge:
Improving the docker-compose page suggesting to create the kong user as a regular db user and not as database administrator. This enforces the least minimum privilege principle and it harden a bit the database suggested configuration.

## Description

Fixes #issue

## Checklist 

### DIFF
--- a/app/_how-tos/gateway/docker-compose.md
+++ b/app/_how-tos/gateway/docker-compose.md
@@ -77,12 +77,12 @@ services:
     image: postgres:latest           # Official Postgres image
     restart: on-failure              # Restart if the container fails
     volumes:
-      - kong_db_data:/var/lib/postgresql  # Mount the volume for persistent data
+      - kong_db_data:/var/lib/postgresql                      # Mount the volume for persistent data
+      - ./init-db.sql:/docker-entrypoint-initdb.d/init-db.sql # Put the initialization SQL script in the current folder
     networks:
       - kong-ee-net                  # Connect to the shared Kong network
     environment:
-      POSTGRES_USER: kong            # Set DB user inside the container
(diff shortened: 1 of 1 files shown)


app/_how-tos/gateway/docker-compose.md
```diff
@@ -132,6 +132,24 @@ EOF
 ```
 <!-- vale on -->
 
+Following the least minimum privilege principle, 'kong' user will be created as a regular user db user.
+
+<!-- vale off -->
+```bash
+cat <<EOF > init-db.sql
+
+CREATE USER kong WITH NOSUPERUSER NOCREATEDB NOCREATEROLE PASSWORD 'kong';
```
mcl-kong, line 141: Will the password for the 'kong' user remain in the sql script? Can the script be removed after the installation?

mcl-kong:
Possibly, this can be further improved by using multiple users with specific privileges: 1 for migrations, 1 for CP, 1 for DP... I got this from @bungle Can you two sync and create a backlog item with this idea?

app/_how-tos/gateway/docker-compose.md
thesp0nge: The password for 'kong' user was first written in the docker compose file. I just moved in the init script because I want the 'kong' user to be created with a smaller subset of permissions.

The script is only used after the first docker compose up. In case the user will remove images, purging the volumes data, the script has to be present.

Guaris:
Let me know after you and @bungle sync so I can review the doc.

thesp0nge:
Do we have any update on this?

PR #5220, opened 2026-05-13
Update expressions.md with beta warning for plugins

ajessup:
Added a warning about the beta status of plugin conditions.

## Description

Fixes #issue

## Preview Links

https://deploy-preview-5220--kongdeveloper.netlify.app/gateway/plugins/expressions/

### DIFF
--- a/app/gateway/plugins/expressions.md
+++ b/app/gateway/plugins/expressions.md
@@ -53,6 +53,9 @@ related_resources:
 
 ---
 
+{:.warning}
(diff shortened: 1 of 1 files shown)


lena-larionova:
We probably want to do this programmatically - there's already a beta label on the page title, so the `beta:true` key should show a banner as well. Merging this as a temp solution.

lena-larionova (approved):

PR #5223, opened 2026-05-13
Fix(Plugins): decK not mentioned as an option in jq plugin example

Guaris:
Tested this locally just in case there was a reason we didn't add it before.


lena-larionova (approved):
I think this was just a copy&paste error. LGTM!

PR #5294, opened 2026-05-19
fix(docs-platform) Trim newlines from codeblock copy function

lena-larionova:
## Description

Removes the extra newline that gets added on codeblocks with the copy button.

Now, instead of this:

```
kongctl get portals

```
The pasted text will look like this:

```
kongctl get portals
```

Issue reported by Rick on Slack.

## Preview Links
Use the copy button on any page with codeblocks, eg https://deploy-preview-5294--kongdeveloper.netlify.app/gateway/get-started/

### DIFF
--- a/app/_assets/javascripts/clipboard_copy.js
+++ b/app/_assets/javascripts/clipboard_copy.js
@@ -1,4 +1,12 @@
 document.addEventListener("clipboard-copy", function (event) {
+  const targetId = event.target.getAttribute("for");
+  if (targetId) {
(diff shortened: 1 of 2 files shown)


fabianrbz:
looks like this is a known bug in shiki, the fix is updating the code before shiki turns it into html
something like https://github.com/shikijs/shiki/issues/3#issuecomment-2272168959
I'll do that

fabianrbz (approved):

PR #5412, opened 2026-06-01
feat(platform): Add tech-preview banner on landing pages

Guaris:
Here is an example of what this would look like, change isn't actually included in the PR only functionality: 
image

```
    columns:
      - blocks:
        - type: card
          config:
            icon: /assets/icons/kubernetes.svg
            title: "Manage {{ site.base_gateway }} in Kubernetes through the Gateway API"
            tech_preview: true
            description: |
              Deploy and configure {{ site.base_gateway }} in Kubernetes using Gateway API resources.
            cta:
              text: Deploy {{ site.base_gateway }} in Kubernetes using Gateway API
              url: /operator/get-started/gateway-api/install/ 

```

### DIFF
--- a/app/_includes/card.html
+++ b/app/_includes/card.html
@@ -43,15 +43,9 @@
     Recommended
     {% endif %}
     </div>
-    {% endif %}
-
-    {% if include.tech_preview %}
+    {% elsif include.tech_preview %}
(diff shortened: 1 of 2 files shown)


lena-larionova:
We should adjust the colour of this banner so it doesn't look exact the same as the recommended banner.

app/_includes/card.html
mheap: Setting `featured: "Tech Preview"` would achieve the same outcome here

lena-larionova (approved):

PR #5432, opened 2026-06-02
docs(mesh): add --skip-verify to universal dp

bartsmykla:
## Description

Recent Kuma releases changed `kuma-dp` to verify the control plane certificate against the system trust store by default. Skipping verification is now opt-in via `--skip-verify` (`KUMA_CONTROL_PLANE_TLS_SKIP_VERIFY`); the production path is to pass the control plane CA with `--ca-cert-file` / `KUMA_CONTROL_PLANE_CA_CERT_FILE`.

The universal getting-started and data-plane-on-universal guides use a self-signed control plane, so `kuma-dp run --cp-address=https://...` now fails verification. This adds `--skip-verify` to those commands in `app/mesh/data-plane-universal.md` and the universal how-tos, with a warning pointing at `--ca-cert-file` for production.

Reviewers: `--skip-verify` is new, so you may want to version-gate these lines for older docs.

## Checklist 

- [ ] Tested how-to docs. If not, note why here. Documentation-only flag change mirroring upstream `kuma-dp` behavior; not run end to end.
- [ ] Add new pages to the product documentation index (if applicable). No new pages.


lmilan:
> Reviewers: --skip-verify is new, so you may want to version-gate these lines for older docs.

@bartsmykla is this new for 2.14?

lmilan (approved):

PR #5441, opened 2026-06-02
fix(docs-improvements): Clean up duplicate vale dictionary entries

lena-larionova:
## Description

In recent PRs, I've noticed that there are a lot of duplicates in our vale dictionary (sometimes up to four copies of a single term). I'm guessing this stems from merge conflict resolutions gone wrong, so this PR cleans that up.

If there were two capitalized/non-capitalized versions of a term, I left only the lowercase one, as Vale will apply that to all capitalizations. 

## Preview Links

N/A


lena-larionova:
All of the copilot comments are incorrect. Vale's case-sensitivity goes in one direction:
* if the dictionary entry is uppercase, it will preserve the case and flag any lowercase versions as incorrect. For example, if the entry is "ACL", "acl" would be considered an error.
* if the dictionary entry is lowercase, it's applied to any version of the word. For example, if the dictionary entry is "acl", both "acl" and "ACL" would be considered correct.

Guaris (approved):

PR #5480, opened 2026-06-05
feat(dcgw): Add OTEL_RESOURCE_ATTRIBUTES as supported

### LINKED ISSUE #5256: [3.15] OTEL_RESOURCE_ATTRIBUTES env var
## Definition of done
- Unless #5197 is complete, update the [DCGW reference with this env var](https://developer.konghq.com/dedicated-cloud-gateways/reference/#kong-gateway-configuration)
- Should be auto added to the kong.conf reference?
- Need to note limitation: "The only validation we are doing is that value is a key value pair with percent encoded in case of special characters as per otel guidelines. And available on the 3.15+. There is no other validation done apart from this."

## Size
S

cloudjumpercat:
## Description

Fixes #5256 

## Preview Links
https://deploy-preview-5480--kongdeveloper.netlify.app/dedicated-cloud-gateways/reference/#kong-gateway-configuration: I manually added the description since this seems to live in [the OTEL schema](https://deploy-preview-5480--kongdeveloper.netlify.app/plugins/opentelemetry/reference/#schema--config-resource-attributes), not kong.conf, but it's still an env var DCGW can specify 🤷‍♀️ 

**_Question for reviewers:_** In 3.15, we're supporting OTEL_RESOURCE_ATTRIBUTES as an env var. This one WON'T have a KONG prefix. DCGW is supporting it, but when I add it to the {% kong_config_table env %} table, it automatically prefixes it with KONG_. I've attempted to update the table with Claude's fix (detailed below). It also seems like the env var isn't in the kong.conf JSON that was just generated a few days ago despite the Jira ticket saying it was completed.

### Claude's fix for the table

This is the summary of what Claude says it's fixing:

**Context**
In Gateway 3.15, Dedicated Cloud Gateways support OTEL_RESOURCE_ATTRIBUTES as an environment variable. Unlike every other DCGW env var, this one is a standard OpenTelemetry variable and does not carry the KONG_ prefix.

The {% kong_config_table env %} block in [app/dedicated-cloud-gateways/reference.md:401-433](vscode-webview://0s4vl89vlnac2q13ughrudaahl5i4rgeef9hk37b89an8sudpmon/app/dedicated-cloud-gateways/reference.md#L401-L433) automatically renders each field name as KONG_<NAME> in env mode. The prefixing happens in format_name at [app/_plugins/drops/kong_config_table.rb:35-40](vscode-webview://0s4vl89vlnac2q13ughrudaahl5i4rgeef9hk37b89an8sudpmon/app/_plugins/drops/kong_config_table.rb#L35-L40):

def format_name(name, mode)
  return name if mode == 'conf'
  return "KONG_#{name.upcase}" if mode == 'env'
  raise "Unknown kong_config_table mode: #{mode}"
end
There is currently no way to render an un-prefixed env var, so otel_resource_attributes renders incorrectly as KONG_OTEL_RESOURCE_ATTRIBUTES.

**Approach**
Add an opt-out flag, prefix: false, that can be set on an individual config entry. When set, env mode renders the upcased name without the KONG_ prefix.

1. app/_plugins/drops/kong_config_table.rb
KongConfigField already holds the per-field config hash in @config, so no signature changes are needed. Update format_name to honor the flag:

def format_name(name, mode)
  return name if mode == 'conf'

  if mode == 'env'
    return name.upcase if @config['prefix'] == false

    return "KONG_#{name.upcase}"
  end

  raise "Unknown kong_config_table mode: #{mode}"
end
Default behavior is unchanged (flag absent or true keeps the KONG_ prefix), so the other 17 files using the block are unaffected.

## Checklist 


app/_plugins/drops/kong_config_table.rb
```diff
@@ -34,7 +34,12 @@ def description
 
         def format_name(name, mode)
           return name if mode == 'conf'
-          return "KONG_#{name.upcase}" if mode == 'env'
+
+          if mode == 'env'
+            return name.upcase if @config['prefix'] == false
+
+            return "KONG_#{name.upcase}"
+          end
```
fabianrbz, line 46: this looks good, @cloudjumpercat can you modify it so that it looks like this?
```ruby
def format_name(name, mode)
  return name if mode == 'conf'

  if mode == 'env'
    if @config['prefix'] == false
      name.upcase
    else
      "KONG_#{name.upcase}"
    end
  else
    raise "Unknown kong_config_table mode: #{mode}"
  end
end
```

lena-larionova:
@cloudjumpercat I had to rebase the base `release/gateway-3.15` branch to pick up Fabian's fixes to netlify builds, make sure to update this one.

lena-larionova (approved):

PR #5520, opened 2026-06-10
feat(ai-gateway): OTEL for AIGW 2.0

jbaross:
## Description

- Policy is required to deploy the plugin
- Min version set to 2.0
- Fixes inconsistent contractions for Open Telemetry in text

Fixes #issue


lena-larionova (changes requested):
app/_data/plugins/otel-metrics.yaml is a data file that filters into both Gateway and AI Gateway metrics docs. It also gets generated into this page: https://developer.konghq.com/gateway/otel-metrics/

You can't change the min_versions in this file directly this way. You'll have to either:
* make a new data file for ai gateway
* make min_version into something like 
```
min_version:
  gateway: '3.14'
  ai_gateway: '2.0'
```
* or create a new section in this file for AI Gateway specific metrics
* maybe something else, if @fabianrbz has a better idea

PR #5532, opened 2026-06-11
fix(gateway): correct markdown backtick

SnowOnion:
## Description

Before:
截屏2026-06-11 09 27 08

After:
截屏2026-06-11 09 28 06

## Checklist 


lena-larionova (approved):
@SnowOnion I'm going to accept this since we're not planning on regenerating 3.14 for now, but this needs to be fixed in source: https://github.com/Kong/kong-ee/blob/master/kong.conf.default#L3378

We generate this document from kong.conf.default, so 3.15 will overwrite this fix. Please fix it there so we can avoid the error in the future.

PR #5556, opened 2026-06-11
chore(gateway): Swap conditional execution how-to guide to use deck

lena-larionova:
## Description

decK supports conditional execution as of 3.14 (slightly after the release date). Pushing this change to 3.15 instead of main because there's been some changes to this guide for conditional execution, so we can publish both updates together.


cloudjumpercat (approved):
LGTM! Tested and it works!

PR #5624, opened 2026-06-18
feat(gateway): Set consumer group ID for kafka/confluent consume

lena-larionova:
## Description

Kafka Consume and Confluent Consume can now set a consumer group ID instead of generating one. This is necessary for brokers that enforce ACLs.

Added plugin examples; I don't think it needs a reference section, but could be convinced.

Examples for https://github.com/Kong/kong-technical-enablement/tree/main/api-gateway/3.15/kafka-consume-consumer-group-id.


tomek-labuk (approved):
Looks good, just a side note: should we rethink our convention of how we frame requirements (instead of 'You have x y z' just state 'x y z' directly)?

lena-larionova:
> Looks good, just a side note: should we rethink our convention of how we frame requirements (instead of 'You have x y z' just state 'x y z' directly)?

Taking this to Slack to decide on a format, and then we can apply it across the board.

PR #5683, opened 2026-06-24
docs(expressions): rename starts_with/ends_with to startsWith/endsWith

chobits:
Renames the expression plugin matcher functions in the docs from snake_case to camelCase:

- `starts_with()` → `startsWith()`
- `ends_with()` → `endsWith()`

Affected files:
- `app/gateway/plugins/expressions.md`
- `app/_includes/gateway/expressions/migrate.md`

We should keep our function name in plugin/expresssions with original [CEL spec](https://github.com/cel-expr/cel-spec/blob/master/doc/langdef.md), and it also aligns with [KEG expression](https://developer.konghq.com/event-gateway/expressions/).

background: https://kongstrong.slack.com/archives/C0A1JBUT7A9/p1782182809676519


bungle (approved):

cloudjumpercat (approved):
LGTM!

PR #5637, opened 2026-06-19
feat(ai-gateway): Add new getting started guide for proxying LLM traffic

tomek-labuk:
## Description

Review comments:
- This is a Konnect-only tutorial
- Make sure it is clear that AI Gateway 2.0 is a Konnect-first product
- For now the tools are konnect-api only, since kongctl support doesn't exist yet
- Following what we have in the current getting started I've included some config descriptions, I might be overkilling them though with too many details
- Do we need UI instructions?
- The includes will be reused across other AI GW 2.0 how-tos
- Do we need some next steps? Like: now that you can proxy LLM traffic, use policies to govern/secure/observe the traffic?

## Preview Links

https://deploy-preview-5637--kongdeveloper.netlify.app/ai-gateway/get-started/

## Checklist 

- [ ] Tested how-to docs. If not, note why here. 
- [ ] All pages contain metadata.
- [ ] Any new docs link to existing docs.
- [ ] All autogenerated instructions render correctly (API, decK, Konnect, Kong Manager).
- [ ] Style guide (capitalized gateway entities, placeholder URLs) implemented correctly.
- [ ] Every page has a `description` entry in frontmatter.
- [ ] Add new pages to the product documentation index (if applicable).

PR #5720, opened 2026-06-25
fix(ai-gateway): v2.0 AI and MCP landing page edits

cloudjumpercat:
## Description

I've made the following edits:
- I've attempted to reduce the redundant descriptions on the AI Gateway landing page. Originally, it had several sections that rehashed the features, entities, etc. in similar wording, just different sections.
- I combined the AI Gateway on Konnect and the "Three traffic types, unified control" sections. The "Three traffic types" section links all said they weren't compatible with on-prem, so I thought that both these sections were talking about Konnect-only features.
- Removed the MCP server and Agent tiles from the get started section. These pointed to reference pages and I couldn't find a how to that would match these. I don't think the reference links will help someone who is looking for a quickstart.
- I consolidated the first h2 subheadings in the MCP landing page because they were saying similar things and we didn't have enough content links to have two ctas tiles.

Additionally, I have the following feedback that would require more edits than I can make here:
- Content on the AI Gateway landing page could be further pared down, there's a lot of content there.
- _Reduce the jargon and descriptions of "how the sausage is made"_ on both the AI Gateway and MCP landing pages: 
  - The pages are currently written in a very feature centric way instead of use case centric. For example, we explain that you can create AI Agent, Model, and MCP server entities and apply policies to them. This would be useful information on a reference page, but someone who is new to AI Gateway likely doesn't care that these are entities. Instead, they want to govern an agent or create a MCP server from their spec. The landing pages should talk more about use cases and then link off to related references where we can then reveal that they achieve their use case via an entity and policy.
  - Compared to competitors, what we are offering seems extremely complex and complicated based on how we are presenting it.
- _Landing pages make on-prem feel unsupported:_ All feature pages (load balancing, streaming, llm metrics) say they aren't compatible with on-prem. Every link you click on (other than cookbooks) seems to be Konnect only, which makes it seem like on-prem isn't supported at all. 
- _Lack of quickstarts/how tos:_ I'm concerned that we don't seem to have any v2.0 quickstarts or how tos other than cookbooks. Almost everything links to a reference, which will make it difficult to use AI Gateway 2.0.
- _Policy pages only contain a reference:_ Is this expected that policy pages only contain the autogenerated reference? This seems like a significant downgrade to the plugin experience where we had examples and overview content about how it worked.

## Preview Links
- /ai-gateway/
- /ai-gateway/mcp/

## Checklist 

### DIFF
--- a/app/_landing_pages/ai-gateway.yaml
+++ b/app/_landing_pages/ai-gateway.yaml
@@ -33,17 +33,12 @@ rows:
               blocks:
                 - type: text
                   text: |
-                    [Sign up for {{site.konnect_short_name}}](https://konghq.com/products/kong-konnect/register?utm_medium=referral&utm_source=docs&utm_content=ai-gateway) to get started with {{site.ai_gateway}}.
-
-                    Or, launch a local demo instance of {{site.ai_gateway}} with a single command:
+                    [Sign up for {{site.konnect_short_name}}](https://konghq.com/products/kong-konnect/register?utm_medium=referral&utm_source=docs&utm_content=ai-gateway) to get started with {{site.ai_gateway}} or launch a local demo instance of {{site.ai_gateway}} with a single command:
                     ```sh
                     curl -Ls https://get.konghq.com/ai | bash
                     ```
 
-                    Or, choose your starting point using one of our quickstart guides:
-                    - Proxy an LLM
-                    - Expose tools via MCP
-                    - Route agents through {{site.ai_gateway}}
+                    Or, proxy your first model through {{site.ai_gateway}} with the [LLM quickstart guide](/ai-gateway/get-started/).
 
       - blocks:
           - type: image
@@ -51,41 +46,11 @@ rows:
               url: /assets/images/gateway/ai-gateway-overview.svg
               alt_text: Overview of AI gateway
 
-
-  - columns:
-      - blocks:
-        - type: card
-          config:
-            title: LLM quickstart
-            description: Proxy your first model through {{site.ai_gateway}} with a guided setup.
-            icon: /assets/icons/llm-quickstart.svg
-            cta:
-              url: /ai-gateway/get-started/
-              align: end
-      - blocks:
-        - type: card
-          config:
-            title: MCP quickstart
-            description: Expose and observe your first tool server over Model Context Protocol.
-            icon: /assets/icons/mcp-quickstart.svg
-            cta:
-              url: /ai-gateway/mcp/
-              align: end
-      - blocks:
-        - type: card
-          config:
-            title: A2A quickstart
-            description: Route and secure agent-to-agent traffic with protocol-aware observability.
-            icon: /assets/icons/a2a-quickstart.svg
-            cta:
-              url: /ai-gateway/a2a/
-              align: end
-
   - header:
       type: h2
       text: "{{site.ai_gateway}} providers"
       description: |
-        {{site.ai_gateway}} routes AI requests through [provider-agnostic APIs](./#universal-api) by combining AI Providers and AI Models.
+        {{site.ai_gateway}} routes AI requests through provider-agnostic APIs by combining AI Providers and AI Models.
         AI Providers store upstream connectivity and credentials, while AI Models reference Providers to expose stable client-facing endpoints and routing behavior.
     column_count: 4
     columns:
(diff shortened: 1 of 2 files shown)


app/_landing_pages/ai-gateway.yaml
```diff
@@ -33,59 +33,24 @@ rows:
...
-                    - Route agents through {{site.ai_gateway}}
+                    Or, proxy your first model through {{site.ai_gateway}} with the [LLM quickstart guide](/ai-gateway/get-started/).
 
       - blocks:
           - type: image
             config:
               url: /assets/images/gateway/ai-gateway-overview.svg
               alt_text: Overview of AI gateway
 
-
```
jbaross, line 54: There will be a real quickstart for each of these (see https://github.com/Kong/developer.konghq.com/pull/5637/changes) so let's restore this section and use links like /ai-gateway/get-started-with-mcp-server/ and /ai-gateway/get-started-with-a2a/

app/_landing_pages/ai-gateway.yaml
jbaross, line 154: Consider hiding these, we stiull don't know if cookbooks will be ready for launch

app/_landing_pages/ai-gateway.yaml
```diff
@@ -153,38 +118,6 @@ rows:
               url: /cookbooks/secure-external-mcp-gateway/
               align: end
 
-  - header:
-      type: h2
-      text: "Deploy {{site.ai_gateway}}"
-    columns:
-      - header:
-          type: h3
-          text: "Tools to manage {{site.ai_gateway}}"
```
jbaross, line 162: The tool links are still valuable but the deployment checklist can be removed

jbaross: alternatively, fold links to the API and kongctl into 'Three traffic types...'

app/_landing_pages/ai-gateway.yaml
```diff
@@ -195,85 +128,23 @@ rows:
...
-                  - [**Load balancing**](/ai-gateway/load-balancing/): Distribute requests across target services for performance and cost efficiency.
+                  {{site.konnect_short_name}} provides a [unified control plane](https://cloud.konghq.com/ai-manager) to create, manage, and monitor LLMs
+                  using the {{site.konnect_short_name}} platform.
 
-                  - [**Retry and fallback**](/ai-gateway/load-balancing/#retry-and-fallback): Route based on performance, cost, or availability.
-
-                  - [**Policy integration**](/ai-gateway/entities/ai-policy/): Attach [AI Policies](/ai-gateway/entities/ai-policy/) for authentication, guardrails, transformations, and governance.
-      - blocks:
+                    Key features include:
+                    * [Routing and load balancing](/ai-gateway/load-balancing/): Configure AI Models and `target_models` routing across [AI Providers](/ai-gateway/entities/ai-provider/).
```
jbaross:
```suggestion
                    * [Routing and load balancing](/ai-gateway/load-balancing/): Route requests across multiple targets per AI Model and across [AI Providers](/ai-gateway/entities/ai-provider/).
```

app/_landing_pages/ai-gateway.yaml
```diff
@@ -294,41 +165,46 @@ rows:
...
+        blocks:
+          - type: structured_text
             config:
-              url: /assets/images/konnect/ai-manager.png
-              alt_text: "{{site.ai_gateway}} Dashboard in Konnect"
-
+              blocks:
+                - type: unordered_list
+                  items:
+                    - "[{{site.ai_gateway}} resource sizing guidelines](/ai-gateway/resource-sizing-guidelines-ai/): Review recommended resource allocation guidelines for {{site.ai_gateway}}."
```
jbaross: I don't think the sizing page makes sense in a Konnect first world, consider removing this link

app/_landing_pages/ai-gateway/mcp.yaml
```diff
@@ -43,35 +43,21 @@ rows:
...
-          - type: structured_text
-            config:
-              header:
-                type: h2
-                text: "Secure and govern MCP servers"
-              blocks:
-                - type: text
-                  text: |
-                    Attach [AI Policies](/ai-gateway/entities/ai-policy/) to your [AI MCP Server](/ai-gateway/entities/ai-mcp-server/) entities to apply security, governance, and observability controls across your MCP infrastructure.
+                    {{site.ai_gateway}} {% new_in 2.0 %} manages MCP traffic through the entity model.
```
jbaross:
```suggestion
                    {{site.ai_gateway}} {% new_in 2.0 %} manages traffic through the [AI MCP Server](/ai-gateway/entities/ai-mcp-server/) entity.
```

jbaross (changes requested):

app/_landing_pages/ai-gateway.yaml
Guaris: `target_models` is now `targets` note to myself

Guaris:
LGTM

PR #5748, opened 2026-06-29
Chore(AIGW): Data Governance review

Guaris:
This PR also removes the Governance H2 from the AIGW landing page since that wasn't really doing anything and makes all the H3's H2's which make the page feel a bit better. 

There weren't many edits on this page, but I did run the claude skill on it.


app/_landing_pages/ai-gateway/ai-data-gov.yaml
```diff
@@ -90,36 +91,36 @@ rows:
     column_count: 2
```
lena-larionova: This would look better as `column_count: 3`, and same with the Observability section.

app/_landing_pages/ai-gateway/ai-data-gov.yaml
lena-larionova: Should this be "User safety"? Applies to the other H2s on the page as well.

PR #5777, opened 2026-06-30
chore(AIGW): Global scope icon in policy info box

lena-larionova:
Adding globe icon and linking to where global scope is defined.

### DIFF
--- a/app/_includes/info_box/sections/scopes.html
+++ b/app/_includes/info_box/sections/scopes.html
@@ -6,7 +6,8 @@
     {% for scope in include.scopes %}
         <div class="flex gap-2 items-center">
             {% if scope == 'global' %}
-                <span class="text-terciary">Global</span>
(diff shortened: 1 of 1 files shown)


app/_includes/info_box/sections/scopes.html
```diff
@@ -6,7 +6,8 @@
     {% for scope in include.scopes %}
         <div class="flex gap-2 items-center">
             {% if scope == 'global' %}
-                <span class="text-terciary">Global</span>
+                {% include mask_image.html image_url='/assets/icons/world.svg' css_classes="w-5 h-5 shrink-0 !bg-icon" %}
```
fabianrbz, line 10: @lena-larionova you'll probably need a wrapping div like the one we have in the `{% else %}`

```
 <div class="flex gap-2">
mask_image
global
```

cloudjumpercat (approved):
LGTM!

PR #5856, opened 2026-07-07
fix(plugin-examples): remove tags if they are not set from plugin

fabianrbz:
## Description

As reported in [slack](https://kongstrong.slack.com/archives/CDSTDSG9J/p1781875708488829)

The terraform manifest contains tags = [] even though it is optional. The APIs for plugins omit the key when tags are not defined, leading to configuration drift when terraform plan is run.

Removing tags from the config if they aren't present.

## Checklist 


cloudjumpercat (approved):
LGTM! Verified with a basic auth example

PR #5867, opened 2026-07-07
Chore(platform): remove segment

Guaris:


cloudjumpercat (approved):

sandhub:
👋🏾 Team, this change has broken all of our tracking and dashboards. This is a blocker for us and would appreciate reverting this change as soon as we can. 

Re: us going over quota for Segment, I am more than happy to chat about ways we can solve this issue. cc: @snehal-kong @Guaris

PR #5910, opened 2026-07-10
feat(gateway): Gateway docker hub & cloudsmith links

lena-larionova:
## Description

Adds links to download locations for gateway packages and images. We've had users comment in the past that they always think the badges in the table are clickable, and expected them to do something, so this addresses that issue.

I chose to link only to the top-level cloudsmith version page instead of specific distros, since they're all available in multiple architectures. I couldn't find a cleaner way to specify them without also hiding that (+ FIPS).

To do: double-check all the links.

(I thought we had a ticket for this, but now I can't find it.)

## Preview Links

https://deploy-preview-5910--kongdeveloper.netlify.app/gateway/version-support-policy/#supported-versions

### DIFF
--- a/app/_data/support/packages.yml
+++ b/app/_data/support/packages.yml
@@ -7,9 +7,11 @@ amazonlinux1:
 amazonlinux2:
   os: Amazon Linux
   version: 2
+  docker_tag: amazonlinux-2
 amazonlinux2023:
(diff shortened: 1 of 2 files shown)


cloudjumpercat (approved):
LGTM!

PR #5934, opened 2026-07-14
chore(aigw): Set canonical urls

lena-larionova:
Setting canonical URLs.
* For pages that have new parallel content, I set the canonical if it didn't have one and removed `status: pending`
* For how-tos that haven't been migrated, I set the related entity/policy/landing page as the canonical and left `status: pending`. Once we migrate those specific how-tos, we should update the URL again and remove `status: pending`.
* Added v1 to links on the old MCP landing page

To check, go to any v1 page (not plugins) and click the link in the banner. You can use https://deploy-preview-5934--kongdeveloper.netlify.app/index/ai-gateway/v1/.


fabianrbz:
@lena-larionova can we point them to an entity page or something? I need to check what happens when they have self referencing canonicals

fabianrbz:
we can't do self-referencing canonical, they probably need to point to `setup a X` section on the entities page

Guaris (approved):

PR #5942, opened 2026-07-14
feat(ai-gateway): Add bedrock how-to for Claude

tomek-labuk:
## Description

Fixes #issue

## Checklist 

### DIFF
--- a/app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
+++ b/app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
@@ -28,8 +28,6 @@ prereqs:
   konnect:
      - name: KONG_NGINX_HTTP_CLIENT_BODY_BUFFER_SIZE=2m
   inline:
-    - title: Congifure kongctl
-      include_content: md/ai-gateway/v2/prereqs/kongctl
     - title: AWS Bedrock
       content: |
         1. Enable model access in AWS Bedrock:
--- a/app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
+++ b/app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
@@ -23,6 +23,10 @@ tools:
   - kongctl
 
 prereqs:
+  gateway:
+    - name: KONG_NGINX_HTTP_CLIENT_BODY_BUFFER_SIZE=2m
+  konnect:
+     - name: KONG_NGINX_HTTP_CLIENT_BODY_BUFFER_SIZE=2m
   inline:
     - title: Congifure kongctl
       include_content: md/ai-gateway/v2/prereqs/kongctl
@@ -60,7 +64,7 @@ tldr:
 
 ## Create an AI Provider entity
 
-Create an [AI Provider](/ai-gateway/entities/ai-provider/) entity to define your connection to AWS Bedrock and store your IAM credentials.
+Create an [AI Provider](/ai-gateway/entities/ai-provider/) entity to define your connection to AWS Bedrock and store your IAM credentials:
 
 ```sh
 kongctl apply -f - --auto-approve --pat "$KONNECT_TOKEN" <<EOF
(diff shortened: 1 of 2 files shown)


app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
```diff
@@ -0,0 +1,225 @@
...
+
+works_on:
+  - konnect
+
+tools:
+  - kongctl
+
+prereqs:
+  inline:
+    - title: Congifure kongctl
```
fabianrbz: we need to remove this prereq, it gets rendered twice because the tool pulls one automatically

app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
```diff
@@ -0,0 +1,225 @@
...
+    capabilities:
+      - generate
+EOF
+```
+
+## Validate the AI Model
+
+Send a test request directly to confirm the setup works before pointing {{ site.claude_code }} at it:
+
+```sh
```
fabianrbz: this curl command fails if the policy is enabled, the error message is:
```
 curl -i -X POST http://localhost:8000/v1/messages \
  -H 'Content-Type: application/json' \
  -H 'anthropic-version: 2023-06-01' \
  --data '{
    "model": "my-claude-bedrock",
    "max_tokens": 1024,
    "messages": [
      {"role": "user", "content": "hello"}
    ]
  }'
HTTP/1.1 500 Internal Server Error
Date: Tue, 14 Jul 2026 14:16:55 GMT
Content-Type: application/json; charset=utf-8
Connection: keep-alive
Content-Length: 55
X-Kong-Response-Latency: 9
Server: kong/2.0.1-ai-gateway
X-Kong-Request-Id: 2fa2d2538feb9c122c23d86f26c67066

{"error":{"message":"failed to get balancer instance"}}%
```

felderi (commented):
**Validation feedback** — I stood up an equivalent config end-to-end today (Claude Code → local DP → Kong → AWS Bedrock, returning 200 with a real Claude Haiku response), so the overall shape is correct 👍. A few refinements:

**1. Namespace: everything lands in `default`, not `ai-gateway-get-started`.** There's no `_defaults.kongctl.namespace` block and the gateway is referenced via `_external`, so every resource this how-to creates ends up in the `default` namespace — not the `ai-gateway-get-started` namespace the kongctl quickstart/prereq sets up. The reason is subtle: child AI Gateway resources (provider/policy/model) inherit their namespace from a **managed** parent gateway, but with `_external` the parent is only *referenced*, not managed, so there's nothing to inherit from → `default`. If you want these grouped with the quickstart flow:
```yaml
_defaults:
  kongctl:
    namespace: ai-gateway-get-started

ai_gateways:
  - ref: ai-quickstart
    name: ai-quickstart
    display_name: "ai-quickstart"
```
(fully declare the gateway instead of `_external`). Verified on a `dev` build: `_external` → `Namespace: default`; full declaration → `Namespace: ai-gateway-get-started` with `_defaults` correctly inherited.

**2. The strip list is thinner than current Claude Code needs.** The policy removes body `[model, output_config]`, and you compensate client-side with `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`. That works, but stripping the full set at the gateway makes it robust without relying on the client env var:
```yaml
remove:
  headers: [anthropic-beta]
  querystring: [beta]
  body: [output_config, context_management, mcp_servers, container, service_tier]
```
That's the set validated against current Claude Code (Bedrock, Azure, and Vertex all reject these Anthropic-only extras). With the fuller strip in place, the `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` step becomes optional rather than required.

**3. Minor.** Prereq title typo: "Congifure kongctl" → "Configure kongctl". Also, for consistency with the Azure sibling doc, consider `model: { name_header: true }` vs this doc's `model: { alias: ... }` — both work, they just differ across the two how-tos.

Nice touch: the "Validate the AI Model" direct-curl step before launching Claude Code is good practice — I've suggested the Azure how-to (#5948) copy it.

felderi (commented):
**Correction to my earlier review (item #1, namespace).** I previously suggested fully declaring the `ai_gateways` block (with `_defaults` namespace) instead of `_external` so the resources land in `ai-gateway-get-started`. **Please disregard that — `_external` is the right call here, and the doc already has it correct.**

The reasoning: the `ai-quickstart` gateway's control plane and data plane are created and owned by the AI quickstart/bootstrap script. kongctl should only manage the **child** resources (provider/policy/model) hanging off it — it should *not* try to own the gateway itself. `_external` is exactly that: reference/link the gateway without taking declarative ownership. Fully declaring it would make kongctl fight the script for the gateway's lifecycle, which we don't want.

The trade-off I flagged is real but expected, not a bug: because child AI Gateway resources inherit their namespace from a *managed* parent, and an `_external` parent isn't managed, the children land in the `default` namespace regardless of `_defaults`. That's an accepted consequence of the current split (script owns the gateway, kongctl owns the children) — the clean fix is a fuller kongctl flow down the road, not taking over the gateway in the doc.

Everything else in my original review stands (the strip-list breadth and the minor typo). Sorry for the churn on this one.

app/_how-tos/ai-gateway/use-claude-code-with-ai-gateway-bedrock.md
```diff
@@ -0,0 +1,236 @@
...
+  - konnect
+
+tools:
+  - kongctl
+
+prereqs:
+  gateway:
+    - name: KONG_NGINX_HTTP_CLIENT_BODY_BUFFER_SIZE=2m
+  konnect:
+     - name: KONG_NGINX_HTTP_CLIENT_BODY_BUFFER_SIZE=2m
```
tomek-labuk: @fabianrbz - we need only Konnect here, but render_gateway_prereq? in prereqs.rb returns false unless products includes 'gateway'. Can we update this somehow?

fabianrbz (approved):

PR #6005, opened 2026-07-17
fix(gateway): Note about StatsD Advanced plugin

lena-larionova:
## Description

Adds note to statsd plugin about parity with statsd advanced, which is not documented, as it was deprecated long before dev site migration.

Issue reported on Slack.

## Preview Links
https://deploy-preview-6005--kongdeveloper.netlify.app/plugins/statsd/


cloudjumpercat (approved):
LGTM!

PR #6008, opened 2026-07-17
chore(aigw): Add version to kongctl prereq

lena-larionova:
## Description

Adds a version requirement to the kongctl prereq, similar to decK. 
For decK, we just tell them it's best to use the latest version; I went with the same approach. We could set a specific version for when AIGW launched on kongctl (1.5.0), but realistically subsequent kongctl versions will continue to have improvements and introduce new resources, etc. So it's better to just keep it up to date.

## Preview Links

https://deploy-preview-6008--kongdeveloper.netlify.app/ai-gateway/get-started/


cloudjumpercat (approved):
LGTM!

PR #6560, opened 2026-08-04
fix(event-gateway): KEG env isolation

### LINKED ISSUE #3412: Rethink env isolation in KEG
## Jobs to be done (optional)
In the KEG docs we have

> Environment isolation: You can create isolated dev, test, and prod namespaces on top of the same physical Kafka cluster. If you have a topic named orders in each virtual cluster, this will map transparently to different backend topics: dev-orders, test-orders, and prod-orders. This provides isolation and automatic name resolution per environment.

However, sharing environment between dev and prod is never good idea and we should not recommend it. What might be a use case is sharing multiple dev environments.

## Definition of done

We do not recommend sharing dev and prod

## Size

S

lena-larionova:
## Description

Rewrite virtual cluster use case for environment isolation example to use multiple dev environments instead of dev, test, and prod, as it's not recommended to share namespaces between dev and prod.

Fixes #3412

### DIFF
--- a/app/event-gateway/entities/virtual-cluster.md
+++ b/app/event-gateway/entities/virtual-cluster.md
@@ -102,11 +102,11 @@ You will need to increase the number of virtual clusters if you want to create m
 
 Here are some common patterns:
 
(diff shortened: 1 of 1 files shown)


cloudjumpercat (approved):
LGTM!

PR #6562, opened 2026-08-04
feat(gateway): Comparison of db-less and decK formats

### LINKED ISSUE #211: Reference for the differences between decK and db-less formats
Source: [go-apiops/deckformat/fileformat.md at main · Kong/go-apiops](https://github.com/Kong/go-apiops/blob/main/deckformat/fileformat.md)

lena-larionova:
## Description

Fixes #211 

Consolidates https://github.com/Kong/go-apiops/blob/main/deckformat/fileformat.md and an existing FAQ from the current decK docs and moves them into their own page.

I considered adding kongctl to this but it's not comparable. While kongctl is a declarative management tool, it doesn't manage any of the same resources as these two.

## Preview Links
https://deploy-preview-6562--kongdeveloper.netlify.app/deck/#kong-gateway-already-has-built-in-declarative-configuration-do-i-still-need-deck
https://deploy-preview-6562--kongdeveloper.netlify.app/deck/dbless-vs-deck/

### DIFF
--- a/app/deck/dbless-vs-deck.md
+++ b/app/deck/dbless-vs-deck.md
@@ -26,6 +26,7 @@ tags:
 
 decK uses its own state file format, which differs from the declarative configuration format built into {{site.base_gateway}} for [DB-less mode](/gateway/db-less-mode/).
 Both formats use YAML and share some metadata fields, but they represent certain entities differently and aren't directly compatible.
+This page explains the differences between the two formats and what to consider when migrating from DB-less config to decK.
 
 ## When to use each format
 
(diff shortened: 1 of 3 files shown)


app/deck/dbless-vs-deck.md
```diff
@@ -0,0 +1,362 @@
...
+      Backup and restore, or configuring {{site.base_gateway}} in DB-less mode. Export with `kong config db_export` and import with `kong config db_import`. 
+  - format: decK
+    usecase: Human-authored, version-controlled {{site.base_gateway}} configuration for a database-backed deployment. Designed for manual editing and GitOps workflows.
+{% endtable %}
+<!--vale on-->
+
+If you're using {{site.base_gateway}} in [DB-less mode](/gateway/db-less-mode/), you can't use decK for `sync`, `dump`, or similar operations because they require write access to the Admin API.
+
+If you're running {{site.base_gateway}} with a database in [traditional](/gateway/traditional-mode/) or in [hybrid mode](/gateway/hybrid-mode/), decK is the better choice for managing entity configuration.
+The `kong config db_import` and `db_export` commands have several limitations:
```
cloudjumpercat:
```suggestion
### {{site.base_gateway}} DB-less format limitations

The `kong config db_import` and `db_export` commands have several limitations:
```
This felt like a shift from the previous content comparing the two.

app/deck/dbless-vs-deck.md
```diff
@@ -0,0 +1,362 @@
...
+Changes aren't propagated to live nodes, so you'd have to restart all nodes manually.
+decK applies changes through the Admin API, so all nodes receive updates automatically.
+* **Deletions**: `db_import` can add and update entities, but it can't delete them.
+It won't remove entities that exist in the database but are absent from the config file.
+* **Direct database access**: `db_import` needs a direct connection to the {{site.base_gateway}} database, which may not be available in production networking environments.
+* **Drift detection**: decK can compare the configuration in {{site.base_gateway}}'s database against your config file and report differences.
+This is useful for CI pipelines or scheduled drift checks.
+* **Readability**: `deck gateway dump` produces a more human-readable file than `db_export`.
+
+decK also has limitations to consider:
```
cloudjumpercat:
```suggestion
### decK limitations

decK also has limitations to consider:
```

cloudjumpercat (changes requested):
I left a few comments, but after reading through the new dbless-vs-deck page, I have some thoughts:

- This page is comparing the two formats, but it's framed like users have a choice and we're helping them decide. I don't think this is accurate. If they are db-less, they use `kong config `. Everyone else uses decK. They don't have a choice. 
- "To convert Consumers and their Consumer Groups from DB-less to decK" _this_ feels like the true purpose of the doc to me. A customer is migrating from DB-less to decK, they need to know how to mentally translate what they were doing with `kong config` to decK. 

If these are both true, I think the page will need to be reworked so it more so follows the migration narrative:
* Keep the [When to use each format](https://deploy-preview-6562--kongdeveloper.netlify.app/deck/dbless-vs-deck/#when-to-use-each-format) intro and table. 
* Leave decK limitations as an h2 for users to consider before migration.
* kong config db_import and db_export command limitations: I think these should move to a [DBless page](https://developer.konghq.com/gateway/db-less-mode/) as something to consider before someone chooses this option.
* All other sections can stay as-is

app/deck/dbless-vs-deck.md
```diff
@@ -0,0 +1,364 @@
+---
+title: "decK state file format vs {{site.base_gateway}} DB-less format"
```
cloudjumpercat:
```suggestion
title: "Migrate from {{site.base_gateway}} DB-less format to decK state file format"
```

cloudjumpercat (approved):
Just one nitpick on the title, this looks great!

PR #3976, opened 2026-01-22
Feat(GW): Adding /api to the dns links to ensure that the API calls work with PrivateLink

smritikjaggi:
Adding /api to the dns links to ensure that the API calls work with PrivateLink

## Description

Fixes #issue

## Checklist 

- [ ] Tested how-to docs. If not, note why here. 
- [ ] All pages contain metadata.
- [ ] Any new docs link to existing docs.
- [ ] All autogenerated instructions render correctly (API, decK, Konnect, Kong Manager).
- [ ] Style guide (capitalized gateway entities, placeholder URLs) implemented correctly.
- [ ] Every page has a `description` entry in frontmatter.
- [ ] Add new pages to the product documentation index (if applicable).

PR #6599, opened 2026-08-07
fix(konnect): Privatelink Konnect API url

lena-larionova:
Replaces https://github.com/Kong/developer.konghq.com/pull/3976. 

Adding note about making konnect api calls via AWS privatelink. This info comes from eng on Slack; verified from original PR.


cloudjumpercat (approved):
LGTM!

PR #6614, opened 2026-08-10
fix(mesh): clarify kumactl usage with Konnect global CP

ashman1984:
## Description

Clarifies when to use `kumactl` and `kubectl` with a Konnect managed global CP and Kubernetes zone CPs.

The current guidance says `kumactl` is read-only because the mesh is deployed on Kubernetes.
That isn't correct for the setup described here, where the global CP is managed by Konnect and Kubernetes hosts the zone CP.

`kumactl` is used to manage mesh resources on the Konnect global CP, while `kubectl` is used to manage Kubernetes resources in the zone cluster.

This updates the shared `kumactl` guidance to make that distinction clear.

## Checklist

- [ ] Tested how-to docs. Not tested end to end.
  This change aligns with the existing [zone migration](https://developer.konghq.com/mesh/migrate-zone/) and [zone federation](https://developer.konghq.com/mesh/federate-zone/) guidance, where `kumactl` is used to apply Mesh resources to the Konnect global CP, and with behaviour we've previously confirmed through Support.


app/_includes/faqs/mesh-configure-kumactl.md
```diff
@@ -1,4 +1,5 @@
-Because the mesh is deployed on Kubernetes, `kumactl` is read-only. You manage resources using `kubectl`. Still, we recommend configuring `kumactl` for visibility and diagnostics.
+When using a global Control Plane managed by {{site.konnect_short_name}}, use `kumactl` to manage Mesh resources there.<br>
+Use `kubectl` to manage Kubernetes resources in the zone cluster. <br>`kumactl` is read-only when connected to a Control Plane running on Kubernetes.
```
lmilan:
```suggestion
When using a global control plane managed by {{site.konnect_short_name}}, use `kumactl` to manage {{site.mesh_product_name}} resources there, and `kubectl` to manage Kubernetes resources in the zone cluster. 

{:.info}
> `kumactl` is read-only when connected to a control plane running on Kubernetes.
```

lmilan (changes requested):

PR #6684, opened 2026-08-14
feat(gateway): How-to guides for routing authenticated clients using datakit

### LINKED ISSUE #5548: Gateway/Datakit: Dynamically set upstream based on authenticating consumer
## Jobs to be done

Create how-to guide to dynamically set upstream based on authenticating consumer.

## Information
Use the OIDC plugin to validate the token to sets a header for the consumer.  Then Datakit reads the header and sets either the service or the upstream.

```
# =============================================================================
# Consumer-based upstream routing with Kong
# =============================================================================
# Goal: for a given caller (identified by a token claim), consistently route to
#       ONE specific backend among several — decided AFTER the route is matched.
#
# Why this shape:
#   Kong's router runs BEFORE auth, so it cannot match on a consumer/claim.
#   Instead we keep ONE route -> ONE service, let OIDC validate the token and
#   set the credential from a claim, then let Datakit read that credential and
#   OVERRIDE the backend in-place. No second route, no header re-routing, no
#   loopback / second proxy hop.
#
# Plugin order (all in the `access` phase) — by static plugin priority, OIDC
# runs before Datakit, so no explicit `ordering` is required:
#   openid-connect  -> validates token, sets credential.id from `client_id` claim
#   datakit         -> reads kong.client.credential, maps id -> backend, sets it
#
# Two backend-write options are shown:
#   VARIANT A (ACTIVE)   : kong.service.upstream  -> named Upstream entity (POOL)
#                          keeps load balancing, health checks, retries.
#   VARIANT B (COMMENTED): kong.service.target    -> single host:port
#                          bypasses LB, health checks, retries. Simpler, no pool.
#   Pick ONE datakit block. They are mutually exclusive on the same route.
#
# Requires: gateway version with DataKit + property/jwt nodes (3.14-era).
#           Confirm the OIDC field name `credential_claim` matches your version.
# =============================================================================

_format_version: "3.0"

# -----------------------------------------------------------------------------
# Backends for VARIANT A (set_upstream). Each Upstream is a POOL of targets, so
# load balancing / health checks / retries are preserved on the override.
# (Not used by VARIANT B — there the host:port comes from the Datakit mapping.)
# -----------------------------------------------------------------------------
upstreams:
  - name: upstream-a
    targets:
      - { target: a1.example:443 }
      - { target: a2.example:443 }   # HA pool

  - name: upstream-b
    targets:
      - { target: b1.example:443 }

  - name: upstream-default          # fallback for unmapped / missing claim
    targets:
      - { target: default.example:443 }

services:
  - name: claims-api
    protocol: https
    host: upstream-default          # default backend; Datakit overrides per-request
    port: 443
    routes:
      - name: claims-route
        paths:
          - /v1/claims

    plugins:
      # -----------------------------------------------------------------------
      # 1) OIDC — validate the bearer token and set the credential from a claim.
      #    `credential_claim` puts the claim value onto credential.id WITHOUT
      #    requiring a Kong Consumer entity (consumer_optional: true).
      # -----------------------------------------------------------------------
      - name: openid-connect
        config:
          issuer: <issuer-url>
          auth_methods:
            - bearer
          credential_claim:          # credential.id := value of this claim
            - client_id
          consumer_optional: true    # no Kong Consumer needed; don't 401 on absence
          # hide_credentials: true   # optional: strip Authorization before upstream

      # -----------------------------------------------------------------------
      # 2) DataKit — VARIANT A (ACTIVE): route to a named UPSTREAM (pool).
      #    Runs after openid-connect by static plugin priority (OIDC's priority is
      #    higher, so it executes first in the access phase; Datakit follows), so
      #    kong.client.credential is already populated. No `ordering` needed. If
      #    you add another access-phase plugin, re-verify with X-DataKit-Debug-Trace.
      #
      #    Flow:  GET_CREDENTIAL -> PICK_UPSTREAM (jq map) -> SET_UPSTREAM
      # -----------------------------------------------------------------------
      - name: datakit
        config:
          # debug: true             # enable only while validating; dev-only output
          nodes:
            # Read the authenticated credential (GET mode: no input connected).
            - name: GET_CREDENTIAL
              type: property
              property: kong.client.credential

            # Map credential.id (the claim value) -> Upstream entity name.
            # Unmapped callers fall through to "upstream-default".
            - name: PICK_UPSTREAM
              type: jq
              input: GET_CREDENTIAL
              jq: |
                {
                  "1234": "upstream-a",
                  "5678": "upstream-b"
                }[.id] // "upstream-default"

            # Apply it (SET mode: input connected). Keeps LB / health / retries.
            - name: SET_UPSTREAM
              type: property
              property: kong.service.upstream
              input: PICK_UPSTREAM

      # -----------------------------------------------------------------------
      # 2) DataKit — VARIANT B (ALTERNATIVE): route to a fixed host:port TARGET.
      #    Bypasses load balancing, health checks, and retries. Use only for
      #    stable single-host backends. To use this instead of Variant A:
      #      - comment out the Variant A `datakit` plugin block above
      #      - uncomment this block
      #      - the `upstreams:` section becomes optional
      #
      # - name: datakit
      #   config:
      #     nodes:
      #       - name: GET_CREDENTIAL
      #         type: property
      #         property: kong.client.credential
      #
      #       # Map credential.id -> "host:port".
      #       - name: PICK_TARGET
      #         type: jq
      #         input: GET_CREDENTIAL
      #         jq: |
      #           {
      #             "1234": "a1.example:443",
      #             "5678": "b1.example:443"
      #           }[.id] // "default.example:443"
      #
      #       # Apply it. NOTE: bypasses balancer (no health checks / retries).
      #       - name: SET_TARGET
      #         type: property
      #         property: kong.service.target
      #         input: PICK_TARGET
      #
      #       # If backends differ on scheme, also set it:
      #       # - name: SET_SCHEME
      #       #   type: property
      #       #   property: kong.service.request.scheme
      #       #   input: <jq node returning "https" | "http">
      # -----------------------------------------------------------------------

# =============================================================================
# Deploy & verify
# =============================================================================
#   deck gateway validate deck.yaml
#   deck gateway diff     deck.yaml
#   deck gateway sync     deck.yaml
#
# Smoke test (trace shows node execution order + each node's output):
#   curl -i https://<proxy>/v1/claims \
#     -H "Authorization: Bearer <token-with-client_id=1234>" \
#     -H "X-DataKit-Debug-Trace: true"
#
# Confirm in the trace:
#   - GET_CREDENTIAL runs AFTER openid-connect and returns the credential
#   - .id carries the claim value (e.g. "1234")
#   - PICK_* resolves to the expected backend (not the default)
#   - SET_* executes before proxying
# =======================================================================
```

## Size
M

lena-larionova:
## Description

Fixes #5548 

Two how-to guides (and parallel plugin examples) that use Datakit with OpenID Connect to route the client based on their auth token, either to a specific Upstream (which provides load balancing and health checking), or directly to a mapped host.

This doesn't use Kong Identity as Kong Identity isn't supported for Datakit _yet_. 
## Preview Links

/how-to/configure-oidc-datakit-host-routing/
/how-to/configure-oidc-datakit-upstream-routing/
/plugins/datakit/
/plugins/datakit/examples/route-upstream-by-credential/
/plugins/datakit/examples/route-host-by-credential/

### DIFF
--- a/app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
+++ b/app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
@@ -81,7 +81,7 @@ This guide shows you how to solve that using two plugins working together in the
 1. [**OpenID Connect**](/plugins/openid-connect/) validates the bearer token and extracts a claim value (such as `client_id`) onto a virtual credential.
 2. [**Datakit**](/plugins/datakit/) reads that credential and maps its value to a named Upstream entity, overriding the backend for that specific request.
 
-All callers share one Route and one Service, and the backend is decided dynamically after authentication.
+All callers share one Route and one Service. After authentication, the backend is selected dynamically based on the client's identity.
 
 This guide uses named Upstream entities, which preserves load balancing, health checks, and retries.
 If you don't need those features and prefer to route directly to a `host:port` without a pool, see [Route requests to different targets based on the authenticated caller](/how-to/configure-oidc-datakit-host-routing/).
--- a/app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
+++ b/app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
@@ -1,0 +1,280 @@
+---
+title: Dynamically set upstream based on the authenticated caller with Datakit
+permalink: /how-to/configure-oidc-datakit-upstream-routing/
+content_type: how_to
+
+related_resources:
+  - text: OpenID Connect plugin
+    url: /plugins/openid-connect/
+  - text: Datakit plugin
+    url: /plugins/datakit/
+  - text: Dynamically set host based on the authenticated caller with Datakit
+    url: /how-to/configure-oidc-datakit-host-routing/
+  - text: OpenID Connect tutorials
+    url: /how-to/?query=openid-connect
+
+plugins:
+  - openid-connect
+  - datakit
+
+entities:
+  - route
+  - service
+  - upstream
+  - plugin
+
+products:
+  - gateway
+
+works_on:
+  - on-prem
+  - konnect
+
+min_version:
+  gateway: '3.14'
+
+tools:
+  - deck
+
+prereqs:
+  entities:
+    services:
+      - example-service
+    routes:
+      - example-route
+  inline:
+    - title: Set up Keycloak
+      include_content: prereqs/auth/oidc/keycloak-oidc-routing
+      icon_url: /assets/icons/keycloak.svg
+
+tags:
+  - authentication
+  - openid-connect
+  - routing
+search_aliases:
+  - oidc
+  - datakit
+  - upstream routing
+
+description: Learn how to validate a bearer token with the OpenID Connect plugin, extract a claim as a credential, then use the Datakit plugin to route the request to a different upstream based on that credential.
+
+tldr:
+  q: How do I route requests to different backends based on who is making the request?
+  a: |
+    Configure the OpenID Connect plugin with `credential_claim` to extract a token claim (for example, `client_id`) and set it as the authenticated credential.
+    Then, configure the Datakit plugin to read that credential, map its value to a named Upstream entity, and override the backend for the request.
+
+cleanup:
+  inline:
+    - title: Clean up Konnect environment
+      include_content: cleanup/platform/konnect
+      icon_url: /assets/icons/gateway.svg
+    - title: Destroy the {{site.base_gateway}} container
+      include_content: cleanup/products/gateway
+      icon_url: /assets/icons/gateway.svg
+automated_tests: false
+---
+
+Kong's router runs before authentication, so it can't directly route traffic based on who is making a request.
+This guide shows you how to solve that using two plugins working together in the `access` phase:
+
+1. [**OpenID Connect**](/plugins/openid-connect/) validates the bearer token and extracts a claim value (such as `client_id`) onto a virtual credential.
+2. [**Datakit**](/plugins/datakit/) reads that credential and maps its value to a named Upstream entity, overriding the backend for that specific request.
+
+All callers share one Route and one Service, and the backend is decided dynamically after authentication.
+
+This guide uses named Upstream entities, which preserves load balancing, health checks, and retries.
+If you don't need those features and prefer to route directly to a `host:port` without a pool, see [Route requests to different targets based on the authenticated caller](/how-to/configure-oidc-datakit-host-routing/).
+
+{:.info}
+> **Note:** The OpenID Connect plugin has a higher [static priority](/gateway/entities/plugin/#plugin-priority) than Datakit, so it always runs first in the `access` phase.
+> No explicit plugin ordering configuration is required.
+
+## Generate salt token
+
+{% include how-tos/steps/deck-salt-token.md %}
+
+## Add upstreams
+
+Create the named Upstream entities that Datakit will route to.
+Each Upstream is a pool of targets, so load balancing, health checks, and retries are preserved when Datakit overrides the backend.
+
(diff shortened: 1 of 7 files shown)


app/_includes/prereqs/auth/oidc/keycloak-oidc-routing.md
```diff
@@ -0,0 +1,145 @@
...
+
+   Because we're using Docker for this demo, we must configure a few networking parameters:
+   * `DECK_ISSUER` and `KEYCLOAK_HOST` use `localhost` because that's how you access Keycloak from your machine.
+   * `DECK_JWKS_ENDPOINT` uses the container name `keycloak` because {{site.base_gateway}} runs inside Docker and reaches Keycloak over the shared `kong-quickstart-net` network.
+
+   In your own setup, especially running outside of a container, you may not need `DECK_JWKS_ENDPOINT`.
+
+1. Open the admin console.
+
+   The default URL is `http://localhost:8080/admin/master/console/`.
```
juliamrch:
```suggestion
1. Open the admin console at default URL `http://localhost:8080/admin/master/console/` and sign in with your username (`admin`) and password (`admin`).
```

app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
```diff
@@ -0,0 +1,280 @@
...
+automated_tests: false
+---
+
+Kong's router runs before authentication, so it can't directly route traffic based on who is making a request.
+This guide shows you how to solve that using two plugins working together in the `access` phase:
+
+1. [**OpenID Connect**](/plugins/openid-connect/) validates the bearer token and extracts a claim value (such as `client_id`) onto a virtual credential.
+2. [**Datakit**](/plugins/datakit/) reads that credential and maps its value to a named Upstream entity, overriding the backend for that specific request.
+
+All callers share one Route and one Service, and the backend is decided dynamically after authentication.
```
juliamrch: decided by who/what?

app/_includes/prereqs/auth/oidc/keycloak-oidc-routing.md
```diff
@@ -0,0 +1,145 @@
+This tutorial requires an identity provider (IdP). If you don't have one, you can use [Keycloak](http://www.keycloak.org/).
+The steps will be similar in other standard identity providers.
+
+For this tutorial, you need three clients: one for the OpenID Connect plugin to use when validating tokens, and two representing the API callers that will be routed to different backends.
+
+#### Install and run Keycloak
+
+1. Install [Keycloak](https://www.keycloak.org/guides) (version 26 or later) on your platform.
+
+   For example, you can use the Keycloak Docker image. The following command attaches Keycloak to the same network as {{site.base_gateway}} so that the OIDC plugin can reach it:
```
juliamrch: Using Docker commands will print logs in the current session and will make it difficult to follow through the rest of the tutorial, to avoid that I suggest this small edit:
 
```suggestion
   For example, you can use the Keycloak Docker image. The following command attaches Keycloak to the same network as {{site.base_gateway}} so that the OIDC plugin can reach it. Open a new session in your terminal and run:
```

juliamrch (approved):
Everything works fine, I suggested small edits to improve the Keycloak setup flow.

app/_how-tos/gateway/configure-oidc-datakit-upstream-routing.md
lena-larionova: By the credentials. I'll rephrase.

lena-larionova:
```suggestion
All callers share one Route and one Service. After authentication, the backend is selected dynamically based on the client's identity.
```

PR #7107, opened 2026-09-02
chore(gateway): OTEL metrics for 3.16

lena-larionova:
## Description

Generate metrics for 3.16, add missing 3.15 metrics.

## Preview Links

https://deploy-preview-7107--kongdeveloper.netlify.app/gateway/otel-metrics/

### DIFF
--- a/app/_data/plugins/otel-metrics.yaml
+++ b/app/_data/plugins/otel-metrics.yaml
@@ -20,7 +20,7 @@ metrics:
       - kong.route.name
       - kong.workspace.name
   - name: kong.latency.internal
-    description: &latency_internal "Kong’s internal processing time in seconds, from when the Gateway receives the request from the client to when it sends the request to the upstream service."
+    description: &latency_internal "Kong's internal processing time in seconds, from when the Gateway receives the request from the client to when it sends the request to the upstream service."
     unit: "s"
     type: Histogram
     attributes:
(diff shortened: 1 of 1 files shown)


app/_data/plugins/otel-metrics.yaml
```diff
@@ -106,11 +144,39 @@ metrics:
       - kong.upstream.state
       - kong.subsystem
   - name: kong.dp.cluster_cert.expiry
-    description: "Timestamp when the data plane’s cluster certificate will expire."
+    description: "Timestamp when the data plane's cluster certificate will expire."
     unit: "s"
     type: Gauge
+  - name: kong.dp.connection.last_seen
+    min_version: "3.16"
+    description: Shows the timestamp when the Control Plane last saw a successful data plane connection. This metric appears only on Control Plane and is reported per Data Plane.
```
cloudjumpercat:
```suggestion
    description: Shows the timestamp when the control plane last saw a successful data plane connection. This metric appears only on control plane and is reported per data plane.
```

app/_data/plugins/otel-metrics.yaml
```diff
@@ -106,11 +144,39 @@ metrics:
...
+    unit: "s"
+    type: Gauge
+    attributes:
+      - kong.dp.node.id
+      - kong.dp.host.name
+      - kong.dp.host.ip
+  - name: kong.dp.config_hash
+    min_version: "3.16"
+    description: Shows the timestamp when the control plane last saw a data plane connect. This metric appears only on control-plane nodes and is reported per data plane.
```
cloudjumpercat:
```suggestion
    description: Shows the timestamp when the control plane last saw a data plane connect. This metric appears only on control plane nodes and is reported per data plane.
```
What's a "control plane node"? Should this be "control planes"?

app/_data/plugins/otel-metrics.yaml
```diff
@@ -106,11 +144,39 @@ metrics:
...
+    unit: "1"
+    type: Gauge
+    attributes:
+      - kong.dp.node.id
+      - kong.dp.host.name
+      - kong.dp.host.ip
+  - name: kong.dp.version.compatible
+    min_version: "3.16"
+    description: "Shows whether a data plane is compatible with the control plane's version. This metric appears only on control-plane nodes and is reported per data plane. A value of 1 means compatible and 0 means not."
```
cloudjumpercat: same comment about control plane nodes

cloudjumpercat:
```suggestion
    description: "Shows whether a data plane is compatible with the control plane's version. This metric appears only on control-plane nodes and is reported per data plane. A value of 1 means compatible and 0 means incompatible."
```

cloudjumpercat (approved):
Left a few suggestions. I also flagged a few "control-plane nodes" since I think this might need to be "control planes" based on context?

app/_data/plugins/otel-metrics.yaml
lena-larionova: Technically in on-prem speak, it's control plane nodes and data plane nodes, but we can standardize.

lena-larionova:
```suggestion
    description: Shows the timestamp when the control plane last saw a data plane connect. This metric appears only on control planes and is reported per data plane node.
```

PR #7129, opened 2026-09-03
fix(support): Title and tldr formatting

lena-larionova:
## Description

When a title has double quotes in it, Algolia doesn't render anything after the first `"`. Double quotes can only surround a title. See https://developer.konghq.com/support/?page=3

Also fixing the missing backtick formatting in the tldr.

Screenshot 2026-09-03 at 1 10 16 PM

Note that you won't be able to preview the title fix; it'll only show up after a re-index.
## Preview Links

/support/error-nginx-kong-lua-9-expected-near-customformat-when-using-a-custom-log-format/


lena-larionova:
Skipping vale on this because it's triggering off of kong.conf properties in the sidebar, which we can't "vale off".

PR #7132, opened 2026-09-04
Feat(aiqw): rate limit adv with creds

### LINKED ISSUE #7080: Support Credential Identifiers in AI Gateway
## Definition of done
Update the policy based rate limiting section of [this doc ](https://developer.konghq.com/ai-gateway/policies/ai-rate-limiting-advanced/#policy-based-rate-limiting) to include the support for credential identifiers in AI Gateway, and include an example. 

## Information
https://konghq.aha.io/features/AI-131
https://konghq.atlassian.net/browse/AG-990

## Size

- S (Small). Example: fixing a broken link, updating wording in an FAQ or paragraph

jbaross:
## Description

See https://konghq.aha.io/features/AI-131 and https://konghq.atlassian.net/browse/AG-990

Fixes #7080 

## Preview Links

- https://deploy-preview-7132--kongdeveloper.netlify.app/ai-gateway/policies/ai-rate-limiting-advanced/

## Checklist 

### DIFF
--- a/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
+++ b/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
@@ -237,6 +237,10 @@ To configure it, specify the function in [`config.request_prompt_count_function`
 
 When using the request prompt provider, it will call the function to get the token count at the request level and implement a limit.
 
+## Identification strategies
+
+The [`config.identifier`](./reference/#schema--config-identifier) field allows you to define an identity based scope for rate limiting. 
+
 ## Known limitations of AI Rate Limiting Advanced
 
 The cost is only reflected during the next request.
--- a/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
+++ b/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
@@ -243,6 +243,30 @@ The AI Rate Limiting Advanced Policy supports several identifiers that allow you
 
 You can configure the type of identifier with the [`config.identifier`](./reference/#schema--config-identifier) field and scope to it using [config.policies.match](./reference/#schema--config-policies-match).
 
+{% table %}
+columns:
+  - title: Type
+    key: type
+  - title: Description
+    key: description
+rows:
+  - type: "`consumer`"
+    description: (Default)
+  - type: "`ip`"
+    description: 
+  - type: "`credential`"
+    description: 
+  - type: "`service`"
+    description: 
+  - type: "`header`"
+    description: 
+  - type: "`path`"
+    description: 
+  - type: "`consumer-group`"
+    description: Requires the policy is attached to an [AI Consumer Group](/ai-gateway/entities/ai-consumer-group/) entity.
+      
+{% endtable %}
+
 The following example modifies the [policy-based rate limiting](/ai-gateway/policies/ai-rate-limiting-advanced/#policy-based-rate-limiting) example to use consumer credentials as an identifier.
 
 {% entity_example %}
--- a/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
+++ b/app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
@@ -239,7 +239,39 @@ When using the request prompt provider, it will call the function to get the tok
 
 ## Identification strategies
 
-The [`config.identifier`](./reference/#schema--config-identifier) field allows you to define an identity based scope for rate limiting. 
+The AI Rate Limiting Advanced Policy supports several identifiers that allow you to define an identity based scope for rate limiting.
+
+You can configure the type of identifier with the [`config.identifier`](./reference/#schema--config-identifier) field and scope to it using [config.policies.match](./reference/#schema--config-policies-match).
+
+The following example modifies the [policy-based rate limiting](/ai-gateway/policies/ai-rate-limiting-advanced/#policy-based-rate-limiting) example to use consumer credentials as an identifier.
+
+{% entity_example %}
+type: policy
+data:
+  display_name: AI Rate Limiting Advanced - Consumer and Model
+  type: ai-rate-limiting-advanced
+  name: ai-rate-limiting-advanced
+  config:
+    identifier: credential
+    policies:
+    - match:
+      - type: credential
+        key: id
+        values:
+          - $CONSUMER_ID
+      - type: model
+        partition_by: true
+        values:
+        - gpt-4o
+      limits:
+        - limit: 100
+          window_size: 60
+        - limit: 1000
+          window_size: 3600
+formats:
+  - kongctl
+  - konnect-api
+{% endentity_example %}
 
 ## Known limitations of AI Rate Limiting Advanced
 


jbaross:
There is no existing documentation of 'virtual credentials' other than an [aside here](https://developer.konghq.com/plugins/ai-mcp-oauth2/#virtual-credentials) which matches the schema description for `credential_claim` but doesn't explain what this is how how to use it.

I think we need a substantial update to [OIDC auth strategies](https://developer.konghq.com/ai-gateway/entities/ai-auth-strategy/#oidc-token-authentication)

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
...
+    key: type
+  - title: Description
+    key: description
+rows:
+  - type: "`consumer`"
+    description: (Default) An [AI Consumer](/ai-gateway/entities/ai-consumer/) entity.
+  - type: "`consumer-group`"
+    description: An [AI Consumer Group](/ai-gateway/entities/ai-consumer-group/) entity. Requires that the Rate Limiting Advanced Policy is referenced by an AI Consumer Group.
+  - type: "`credential`"
+    description: A virtual credential created by an [OIDC AI Auth Strategy](/ai-gateway/entities/ai-auth-strategy/#oidc-token-authentication).
```
cloudjumpercat:
```suggestion
    description: A virtual credential derived by an [OIDC AI Auth Strategy](/ai-gateway/entities/ai-auth-strategy/#oidc-token-authentication) from the `credential_claim` field.
```

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
 
 When using the request prompt provider, it will call the function to get the token count at the request level and implement a limit.
 
+## Identification strategies
```
cloudjumpercat:
```suggestion
## Incrementing counters per identity
```

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
...
+
+The AI Rate Limiting Advanced policy supports several identifiers that allow you to define an identity based scopes for rate limiting.
```
cloudjumpercat, line 242:
```suggestion
The AI Rate Limiting Advanced policy supports several identifiers that allow you to define an how rate limiting counters are incremented.
```
"scopes" could be misleading since this isn't scoping the policy to the identity, that is done with the actual scope. This is about how counters are incremented.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
...
+
+You can configure the type of identifier with the [`config.identifier`](./reference/#schema--config-identifier) field and scope to it using [`config.policies.match`](./reference/#schema--config-policies-match). Counters update based on the scope you set and increment separately for each unique identifier.
```
cloudjumpercat, line 244:
```suggestion
You have two options to control how counters increment:
* **Identifier:** [`config.identifier`](./reference/#schema--config-identifier) accepts set strings, such as `consumer` or `ip`, and increments the counter based on these. For example, two AI Consumers in the group sending requests from the same IP share a counter. 
* **Policy match:** [`config.policies.match`](./reference/#schema--config-policies-match) allows you to configure more complex counter increments. This array lets a single policy instance assign different numeric limits to different matched values. For example, 100/min for Consumer A, 1000/hr for model `gpt-4o`. 
```
Based on my research, this isn't quite true. These are both incrementing counters, but identifier accepts only a string whereas `policies.match` is an array that allows for a more granular approach. I'd recommend verifying my feedback, but this suggestion is based on my understanding.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
...
+    description: An [AI Consumer Group](/ai-gateway/entities/ai-consumer-group/) entity. Requires that the Rate Limiting Advanced Policy is referenced by an AI Consumer Group.
+  - type: "`credential`"
+    description: A virtual credential created by an [OIDC AI Auth Strategy](/ai-gateway/entities/ai-auth-strategy/#oidc-token-authentication).
+  - type: "`header`"
+    description: An HTTP header.
+  - type: "`ip`"
+    description: An `ip` address.
+  - type: "`path`"
+    description: An endpoint URL path.
+{% endtable %}
```
cloudjumpercat, line 265:
```suggestion
```
Recommending we remove this table since this is mostly restating what's already in the schema. I'm going to pull out the virtual credentials content though into it's own spot.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
```diff
@@ -237,6 +237,33 @@ To configure it, specify the function in [`config.request_prompt_count_function`
...
+
```
cloudjumpercat, line 266:
```suggestion

Both [`config.identifier`](./reference/#schema--config-identifier) and [`config.policies.match`](./reference/#schema--config-policies-match) allow you to increment counters per `credential`.
This is a virtual credential derived by an [OIDC AI Auth Strategy](/ai-gateway/entities/ai-auth-strategy/#oidc-token-authentication) from the `credential_claim` field.
```

cloudjumpercat (changes requested):
Left some suggestions! Mainly, based on my research, I tried to remove "scopes" type language because these fields seem to be controlling counters and I didn't want to confuse anyone who think that these might actually control how a policy is scoped. I also suggested removing the table since it's mostly repeating the information that's already in the schema. I did pull out the virtual credential content from the table.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
jbaross: My understanding based on the [AHA ticket](https://konghq.aha.io/features/AI-131) was that both were needed in conjunction, but this is very sparse. Happy to update but would like some links to whatever other information sources you have found

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
jbaross:
```suggestion
The AI Rate Limiting Advanced policy supports several identifiers that allow you to define which identifiers are used to track rate limiting counters.
```

How about this?

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
jbaross: again, happy to add this but would like to know how you discovered the info

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
jbaross: I think we should keep this table, the schema list the allowed values but does not describe them. 

These options are not self-explanatory, similar to [token count strategies](https://developer.konghq.com/ai-gateway/policies/ai-rate-limiting-advanced/#token-count-strategies) on the same page or the list of [fields](https://developer.konghq.com/ai-gateway/policies/ai-sanitizer/#fields-that-can-be-anonymized) for PII sanitiser this table is additive to the schema

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
cloudjumpercat: It was both my understanding and research with Claude. 

I knew you could already scope the original plugin to Consumers/Consumer groups, so it didn't make sense that identifier and policies.match would do the same thing (scope to an entity). So I knew they must be doing something on top of the plugin scope. I asked Claude what identifier did if it's not the scope, and it's response + what the schema said lead me to understand that it's telling what the counters should increment by. 

I do think I got policies.match wrong though. I think this is a more granular RL trigger in that no matter the scope of the plugin/policy, this can further help you define limits. So I could scope the plugin/policy to Consumers, so it is only triggered on Consumers. But, I could use policies.match to specify that Consumer Group A gets 100 requests/sec while Consumer Group B gets 50 requests/sec. That policies.match
 field feels a lot more flexible in what it can do. And based on the 3.14 release note, it can also do counter capabilities: "Added a policy-based rate limiting mode with multi-dimensional match conditions and request counting strategy." 

There is nothing in the schema that says you need to use both identifier and policies.match . If they need to be used together, that's something the schema needs to call out and then the docs can also call it out. I think people who just want a simple way to configure the counter increments would use identifier. People who want more granular control, including of the increment counter but not only that, would use policies.match . Since policies.match was introduced in 3.14, it's newer but doesn't replace indentifier.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
cloudjumpercat: I'm okay to leave the table.

app/_ai_gateway_policies/ai-rate-limiting-advanced/index.md
cloudjumpercat: I still don't think they are tracking the counters. They are telling the counters when they need to increment. Like if you increment by IP, this tells RLA that every time it sees a new IP to increment the counter. Or if there is the same IP, don't increment the counter.

And based on https://github.com/Kong/developer.konghq.com/pull/7132#discussion_r4047540824, `policies.match` is doing more than just counters now that I look at it again. It can do more granular counters, but it's also doing more I think. We might need a human or Hal to chime in here because the schema descriptions are confusing.

jbaross: I agree that setting `config.identifier` 

> tells RLA that every time it sees a new IP to increment the counter

However, I think my description comunicates this better. The salient fact is that you're making the choice use ips rather than e.g. credentials to determine what the buckets are for the rates. In other words to 'track' ips not credentials. We're not changing anything about how those buckets are incremented.

jbaross: For `policies.match` I asked for input here https://kongstrong.slack.com/archives/C0AKJ4F2LBS/p1789987030812819 and hal's input here https://kongstrong.slack.com/archives/C06GB3KM155/p1789987568843349?thread_ts=1789987560.997179&cid=C06GB3KM155 

per Hal identifier sets the base dimension used to build the rate limit counter key and match decides which limits to apply. I think adopting this language may help

PR #7161, opened 2026-09-08
Feat(platform)How-to skill 

Guaris:
a skill to help us go faster when creating how-tos.

### DIFF
--- a/.claude/skills/how-to-starter/SKILL.md
+++ b/.claude/skills/how-to-starter/SKILL.md
@@ -1,0 +1,39 @@
+---
+name: how-to-starter
+description: >
+  Scaffold a brand-new how-to guide for developer.konghq.com from a use case description
+  plus whatever source material a writer has (PRD, API/OpenAPI spec, Aha ticket, raw
+  config, curl). Use whenever someone asks to start, scaffold, bootstrap, or generate the
+  skeleton of a new how-to page. Produces structure only, not finished prose: real
+  frontmatter filled in where known and `<!-- TODO -->` where not, terse one-line
+  placeholder steps, the right tool choice (kongctl/deck/admin-api), and the right
+  platform tags (`{% entity_examples %}`, `{% validation %}`), always ending in a
+  Validate step. The writer adds the real words and tests the doc themselves. If the use
+  case is Kong Identity, DCGW Terraform, or a third-party plugin, prefer
+  kong-identity-how-to, dcgw-terraform-how-to, or third-party-plugin instead — those
+  write full prose for their domains. Not for revising an existing how-to.
+---
+
+Read `references/scaffold-patterns.md` before doing anything else — it has the frontmatter
+template, tag syntax, and placeholder conventions.
+
+1. Ask what the how-to should accomplish, and for any source material available. Read
+   whatever's provided.
+2. Derive `products`, `works_on`, `tools`, and `entities` from that material rather than
+   asking — prefer `kongctl`/`deck` over `admin-api` when either would work. Only ask for
+   what you truly can't infer, like the permalink slug or `min_version`.
+3. Check 2-3 sibling files in the target `app/_how-tos/<product>/` directory to match that
+   product's real permalink and frontmatter conventions before writing anything.
+4. Write the file. Fill every frontmatter field you know; drop an inline
+   `<!-- TODO: ... -->` for anything unknown — never guess a value.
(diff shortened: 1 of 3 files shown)


lena-larionova:
We should test some scenarios where you pass a list of steps to it. For example:

1. Set up a consumer with auth
2. Use xyz plugin for abc
3. do another thing
4. validate

.claude/skills/how-to-starter/references/scaffold-patterns.md
```diff
@@ -0,0 +1,191 @@
...
+
+plugins:
+  - <plugin slug, if this how-to is about a specific plugin>
+
+tags:
+  - <slug>
+
+tldr:
+  q: How do I <accomplish the use case>?
+  a: <One or two sentences, the shape of the answer>
```
cloudjumpercat, line 50:
```suggestion
  a: <One or two sentences that describes how to accomplish the task in the how to for advanced users who don't want to follow the full how to. This is NOT an overview of the guide.>
```
Just because I saw this behavior in another PR, Claude thought we wanted like a guide overview here, not a "cheat code" for advanced users.

.claude/skills/how-to-starter/references/scaffold-patterns.md
```diff
@@ -0,0 +1,191 @@
...
+
+related_resources:
+  - text: <link text>
+    url: <path>
+
+prereqs:
```
cloudjumpercat, line 56: I don't remember all the different iterations of what you can do in the prereqs, but I think we need them here because what we have here is currently a small subset. I'd just use Claude to crawl the how tos and add more options here.

.claude/skills/how-to-starter/references/scaffold-patterns.md
```diff
@@ -0,0 +1,191 @@
...
+Each `## H2` in a how-to body becomes a rendered, collapsible step. Order:
+
+```
+## <First config step>
+## <Next config step>
+...
+## Validate          <- always last, always present
+```
+
+## `entity_examples` vs `entity_example`
```
cloudjumpercat, line 103: Did you intentionally omit Terraform examples and konnect_api_request? Some Konnect features do not have kongctl or TF support on release, so I'll write API how tos. I wasn't sure if you wanted this to be comprehensive of all block "types" or to favor the ones we want to favor, like kongctl, etc.

.claude/skills/how-to-starter/SKILL.md
```diff
@@ -0,0 +1,44 @@
+---
+name: how-to-starter
+description: >
+  Scaffold a brand-new how-to guide for developer.konghq.com from a use case description
+  plus whatever source material a writer has (PRD, API/OpenAPI spec, Aha ticket, raw
+  config, curl). Use whenever someone asks to start, scaffold, bootstrap, or generate the
+  skeleton of a new how-to page. Produces structure only, not finished prose: real
```
cloudjumpercat, line 7: Why not finished prose? Would that be a separate skill?

.claude/skills/how-to-starter/SKILL.md
```diff
@@ -0,0 +1,44 @@
...
+template, tag syntax, and placeholder conventions.
+
+1. Ask what the how-to should accomplish, and for any source material available. Read
+   whatever's provided.
+2. Derive `products`, `works_on`, `tools`, and `entities` from that material rather than
+   asking — prefer `kongctl`/`deck` over `admin-api` when either would work. Only ask for
+   what you truly can't infer, like the permalink slug or `min_version`.
+3. Check 2-3 sibling files in the target `app/_how-tos/<product>/` directory to match that
+   product's real permalink and frontmatter conventions before writing anything.
+4. Write the file. Fill every frontmatter field you know; drop an inline
```
cloudjumpercat, line 35: How does this derive the correct format of the steps? As in, does it just guess based on the use case if a user needs to create an AI Model, auth strategy, AI Consumer, Kong Identity auth server? 

When I've drafted how tos with a skill before, I try to give it an overview of what I think will be needed and the rough order, so 
1. Create the auth server/clients/etc.
2. Create the AI Consumer with oauth client credentials
3. Configure the AI Model 

Otherwise it struggles with the flow sometimes. It will critique my rough order and point out what it thinks is wrong, but I think the outline gives it something to go on.

cloudjumpercat (commented):
Left mostly questions!

fabianrbz (approved):

PR #7278, opened 2026-09-17
Updated to zero downtime (ish) updates for KIC to KO

justindavies:
fix(kong-operator): Update the KIC docs to reduce interruption during mogration

## Description

Fixes #issue

## Checklist 


app/operator/migrate/migrate-kic-to-ko-gateway-api.md
```diff
@@ -0,0 +1,356 @@
...
+
+```bash
+kubectl wait -n "${NAMESPACE}" --for=condition=Programmed --timeout=300s gateway/kong-migrated
+```
+
+The replacement is ready for you to attach routes.
+
+## Attach routes to both Gateways
+
+An HTTPRoute is attached to a Gateway through `spec.parentRefs`. Add the replacement Gateway as another parent so both installations can serve the same route during validation.
```
randmonkey, line 192: Could we mention that other kind of routes can also be transferred (But with different versions of KO)?

app/operator/migrate/migrate-kic-to-ko-gateway-api.md
```diff
@@ -0,0 +1,356 @@
...
+For example, append `kong-migrated` to an existing route named `example-route`:
+
+```bash
+kubectl patch httproute example-route -n "${NAMESPACE}" --type json \
+  -p '[{"op":"add","path":"/spec/parentRefs/-","value":{"kind":"Gateway","name":"kong-migrated"}}]'
+```
+
+This preserves the existing parent references. The route must already have a `parentRefs` list; check that `kong-migrated` is not already listed before applying or retrying the patch. If the old attachment uses `sectionName` or `port`, include the appropriate listener restriction in the new reference as well.
+
+Repeat for every HTTPRoute attached to the old Gateway that you want to migrate. Confirm each route references both Gateways before validating the replacement. Keep the old parent references until cleanup.
```
randmonkey, line 203: Also could we use `route` instead of `HTTPRoute` to be inclusive of other kinds of routes, like `TLSRoute`, `GRPCRoute`?

PR #7303, opened 2026-09-19
feat(event-gateway): Request rule validator policy

lena-larionova:
## Description

New event gateway policy in 1.3: request rule validator. This adds the policy ref doc and an example.

Note: can't test this yet, the API isn't updated in Konnect.

## Preview Links

/event-gateway/policies/request-rule-validator/


fabianrbz (approved):

app/_event_gateway_policies/request-rule-validator/examples/enforce-topic-conventions.yml
```diff
@@ -0,0 +1,46 @@
...
+weight: 900
+
+name: request_rule_validator_policy
+type: request_rule_validator
+config:
+  requests:
+    - type: create_topics
+      rules:
+        - description: "Topic name must start with \"team-a.\""
+          rule: '{{topic.name.startsWith("team-a.")}}'
```
jakubdyszkiewicz, line 28: I'd drop `{{`, `}}` for now.

AFAIK `{{` means that we use new CEL expressions while not putting `{{` means it uses regular JS-like syntax used for KEG in other places.

app/_event_gateway_policies/request-rule-validator/index.md
```diff
@@ -0,0 +1,138 @@
...
+Use it to enforce conventions that the [ACLs policy](/event-gateway/policies/acl/) can't express, because ACLs decide whether a principal can perform an action, not what values the action carries.
+
+For example, you can use the Request Rule Validator policy to:
+* Limit the number of partitions or replicas a topic can have.
+* Enforce a naming convention for topics or consumer groups.
+* Require producers to acknowledge writes from all in-sync replicas.
+* Block changes to broker-level configuration.
+* Prevent overly permissive ACL bindings, like granting `ALL` or using a wildcard principal.
+
+Each rule describes the valid state of a request as a [CEL expression](/event-gateway/expressions/). When the expression evaluates to `false`, the rule's configured action runs.
```
jakubdyszkiewicz, line 54:
```suggestion
Each rule describes the valid state of a request as a [gateway expression](/event-gateway/expressions/). When the expression evaluates to `false`, the rule's configured action runs.
```

remove CEL for now

app/_event_gateway_policies/request-rule-validator/index.md
```diff
@@ -0,0 +1,138 @@
...
+  egw-x client: POLICY_VIOLATION
+  end
+
+{% endmermaid %}
+<!--vale on-->
+
+### Request types and available values
+
+You configure rules per Kafka request type.
+Each request type exposes a different set of values to the rule's CEL expression, matching the fields of that request.
```
jakubdyszkiewicz, line 109:
```suggestion
Each request type exposes a different set of values to the rule's expression, matching the fields of that request.
```

app/_event_gateway_policies/request-rule-validator/index.md
```diff
@@ -0,0 +1,138 @@
...
+Some request types, like `create_topics`, carry a batch of items in a single request.
+{{site.event_gateway_short}} evaluates every rule against each item independently.
+When one item in a batch violates a `reject` rule, only that item is rejected with `POLICY_VIOLATION`, while the rest of the batch proceeds normally.
+
+### Actions
+
+Each rule configures one of two actions, which runs when the rule's expression evaluates to `false`:
+
+* `reject`: Fails the request, or the offending item in a batch, with the `POLICY_VIOLATION` error code.
+* `passthrough`: Lets the request continue, but records the violation the same way `reject` does.
```
jakubdyszkiewicz, line 125:
```suggestion
* `passthrough`: Lets the request continue, but records the violation the same way `reject` does (metrics and logs on a debug level).
```

jakubdyszkiewicz (approved):
some nits, but looks good overall

I think we need more examples. This feature exposes quite a lot of things but we only put one example. I'll try to contribute it on followup PRs

app/_event_gateway_policies/request-rule-validator/examples/enforce-topic-conventions.yml
fabianrbz: I've tried this, it only works if I remove it from `rule: '{{topic.name.startsWith("team-a.")}}'` but we need them for the others

fabianrbz: NOTE: this was done with CEL, which will be the new expression language going forward. that's why it's using `{{ }}`. 1.3 should land after CEL so this should be ok

